PR #3, opened 2018-10-26
Studio support in the client 

Alexis-Jacob:


deepoctl/cmds/studio_helpers/image.py
```diff
@@ -0,0 +1,59 @@
...
+        else:
+            raise RuntimeError("Can't find the dataset {}".format(dataset_name))
+
+        data = {
+            "task": "import_images",
+            "org_slug": org_slug,
+            "cuuid": commit_pk,
+            "dataset_name": dataset_name
+        }
+        if type(path) is not list:
```
maingoh: isinstance ?

deepoctl/cmds/studio_helpers/task.py
```diff
@@ -0,0 +1,22 @@
...
+class Task(object):
+    def __init__(self, helper, pk=None, files=None):
+        self._helper = helper
+
+    def retrieve(self, task_id, wait=True):
+        print(task_id)
+        ret = self._helper.get('manage/tasks/{}/'.format(task_id))
+        if not wait:
+            return ret
+        while ret['next'] != None and ret['status'] != 'SUCCESS':
```
maingoh: Ideally you should use exp backoff and jiter https://github.com/Deepomatic/deepomatic-client-python/blob/ef579ac78bae3fae00c196bc2fae59876c06a29b/deepomatic/resources/task.py#L65 or the API might suffer

deepoctl/cmds/studio_helpers/image.py
```diff
@@ -0,0 +1,59 @@
...
+            path = [path]
+        if type(path) is list:
+            files = []
+            for elem in path:
+                for file in get_files(elem, recursive):
+                    files.append(('files', open(file, 'rb')))
```
maingoh: it's better to close the files explicitely

deepoctl/cmds/studio_helpers/http_helper.py
```diff
@@ -0,0 +1,200 @@
+# -*- coding: utf-8 -*-
+import os
+import sys
+import requests
+import platform
+import json
+from six import string_types
+from requests.structures import CaseInsensitiveDict
+
+class HTTPHelper(object):
```
maingoh: We should find a way to use the same as deepomatic-client-python (which I updated this week)

deepoctl/cmds/studio_helpers/http_helper.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def make_request(self, func, resource, params=None, data=None, content_type='application/json', files=None, stream=False, *args, **kwargs):
+        if isinstance(data, dict) or isinstance(data, list):
+            if content_type is not None:
+                if content_type.strip() == 'application/json':
+                    data = json.dumps(data)
+                elif content_type.strip() == 'multipart/mixed':
+                    content_type = None  # will be automatically set to multipart
+                    data, files = self.dump_json_for_multipart(data, files)
+                elif content_type.strip() == 'multipart/form':
```
maingoh: `multipart/mixed` should be enough no ?

deepoctl/cmds/studio_helpers/image.py
```diff
@@ -0,0 +1,59 @@
...
+def get_files(path, recursive=True):
+    if os.path.isfile(path):
+        if path.split('.')[-1].lower() in supported_image_formats:
+            yield path
+    elif os.path.isdir(path) and recursive:
+        for root, dirs, files in os.walk(path, topdown=False):
+            for name in files:
+                if name.split('.')[-1].lower() in supported_image_formats:
+                    yield os.path.join(root, name)
+    elif os.path.isdir(path) and not recursive:
```
maingoh: you could have one isdir if

deepoctl/cmds/studio_helpers/http_helper.py
Alexis-Jacob: no it's not working when I tried that

deepoctl/cmds/studio_helpers/image.py
Alexis-Jacob: true

deepoctl/cmds/studio_helpers/http_helper.py
Alexis-Jacob: in another pr no ?

deepoctl/cmds/studio_helpers/image.py
Alexis-Jacob: it was done in the http_helper but I agree, maybe change that when `studio_helpers` is merged with `deepomatic-client` ?

maingoh: Yeah it's not a big deal, just things we should think about later

deepoctl/cmds/studio_helpers/http_helper.py
maingoh: Yes

deepoctl/cmds/studio_helpers/http_helper.py
maingoh: Hmm, weird, we will upgrade the http helper from deepomatic-client-python and try again. It should work. Or probably there is a different behavior on studio side.

PR #12, opened 2018-11-20
Fix directory output data

alemoreau:
In order to implement the **directory output** , a frame name is generated and attached to each input frame. It also enables to print where the output is currently written.
There is now a progressbar and overall better logging.

### DIFF
--- a/deepoctl/io_data.py
+++ b/deepoctl/io_data.py
@@ -4,6 +4,7 @@ import json
 import imutils
 import logging
 import cv2
+import time
 import threading
 from progressbar import UnknownLength, ProgressBar
 
@@ -12,6 +13,7 @@ try:
 except ImportError:
     from queue import Queue, LifoQueue, Empty
 
+INPUT_QUEUE_MAX_SIZE = 50
 
 def print_log(log):
     sys.stdout.write("\033[F") 
@@ -88,7 +90,19 @@ def input_loop(kwargs, worker_thread):
                             input_queue.task_done()
                         except Empty:
                             break
+
+                while input_queue.qsize() > INPUT_QUEUE_MAX_SIZE:
+                    time.sleep(1)
+
                 input_queue.put(frame)
+            # notify worker_thread that input stream is over
+            input_queue.put(None)
+
+            while worker.isAlive():
+                worker.join(1)
+            while output_thread.isAlive():
+                output_thread.join(1)
+
         except KeyboardInterrupt:
             logging.info('Stopping input')
             while not input_queue.empty():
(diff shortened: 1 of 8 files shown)


deepoctl/io_data.py
```diff
@@ -2,32 +2,41 @@
...
+    print(log)
+
+def get_input(descriptor, args):
     if (descriptor is None):
         raise NameError('No input specified. use -i flag')
     elif os.path.exists(descriptor):
         if os.path.isfile(descriptor):
             if ImageInputData.is_valid(descriptor):
-                return ImageInputData(descriptor)
+                return ImageInputData(descriptor, **args)
```
maingoh: if this is a dictionnary, call it kwargs

deepoctl/io_data.py
```diff
@@ -113,7 +135,13 @@ def next(self):
     def get_fps(self):
         raise NotImplementedError
 
-    def get_frame_number(self):
+    def get_frame_name(self):
+        raise NotImplementedError
```
maingoh: `()`

deepoctl/io_data.py
```diff
@@ -161,32 +196,42 @@ def is_valid(cls, descriptor):
...
+            return self._cap.get(cv2.CAP_PROP_FPS)
 
-    def get_frame_number(self):
-        return self._cap.get(cv2.CAP_PROP_FRAME_COUNT)
+    def get_frame_index(self):
+        return self._i
+
+    def get_frame_count(self):
+        if (self._cap is not None):
+            return self._cap.get(cv2.CAP_PROP_FRAME_COUNT)
```
maingoh: `return None` otherwise

deepoctl/workflow_abstraction.py
```diff
@@ -105,8 +105,11 @@ def __init__(self, recognition_version_id, amqp_url, routing_key):
                 recognition_version_id = int(recognition_version_id)
                 self._recognition = self._client.create_recognition(version_id=recognition_version_id)
             except ValueError:
-                logging.warning("Cannot cast recognition ID into a number")
+                logging.error("Cannot cast recognition ID into a number")
```
maingoh: Is it something that can happen ? Don't we want to make let it raise ?

deepoctl/workflow_abstraction.py
```diff
@@ -105,8 +105,11 @@ def __init__(self, recognition_version_id, amqp_url, routing_key):
...
+            except:
```
maingoh: - In pep8 you need to do `except Exception:`
- Is it something that can happen ? Don't we want to make let it raise ? Or print the traceback ?

PR #22, opened 2019-01-25
Use new deepomatic-api 0.8.0 and optional deepomatic-rpc 0.7.5

thomas-riccardi:


thomas-riccardi:
We should in fact split `deepocli/workflow_abstraction.py`: one file per implementation.
It would avoid names collisions.

PR #25, opened 2019-01-30
Update to new deepo rpc interface

alemoreau:
- Updated code for deepomatic-rpc 0.8.0, tested in local. 
- Splitted workflows into multiple files


deepocli/workflow_abstraction.py
```diff
@@ -101,13 +104,15 @@ def __init__(self, recognition_version_id, amqp_url, routing_key):
...
 
         if can_use_rpc:
-            self._client = deepomatic.rpc.client.Client(self._amqp_url, max_callback=4)
+            self._client = deepomatic.rpc.client.Client(self._amqp_url)
             self._recognition = None
             try:
                 recognition_version_id = int(recognition_version_id)
-                self._recognition = self._client.create_recognition(version_id=recognition_version_id)
+                self._command_mix = create_recognition_command_mix(recognition_version_id)
+                command_queue, response_queue, self._consumer = self._client.add_rpc_queues(routing_key, response_queue_name=self._response_routing_key)
```
maingoh: I added in the README later, it's probably better to purge the queue before you use it. Since now it is always the same response queue, you might want to get ride of older messages (for example if you did a ctrl+c at the previous launch):
```
response_queue.purge()
```

deepocli/workflow_abstraction.py
```diff
@@ -141,4 +147,4 @@ def get_workflow(args):
         workflow = CloudRecognition(recognition_id)
     else:
         return None
-    return workflow
+    return workflow
```
maingoh: newline

deepocli/workflow_abstraction.py
maingoh: I hesitated but by experience it has always been a problem to use temporary queues. Now that we have a robust amqp client it might be okay but using the same queue look much simpler.

thomas-riccardi: One unique queue breaks as soon as there are multiple instances of deepocli working with the same rabbitmq. It will happen in the near future.

maingoh: Okay, I improved the client interface to support both

PR #33, opened 2019-02-28
change 'deepo feedback' to 'deepo studio add_images'

alemoreau:
to validate https://airtable.com/tbl0NdAjudtx4qYPZ/viway1kb3CzTjERth/recVBXcgVrV0cEPma?blocks=hide


alemoreau:
Should I leave the old `deepo feedback` command available ?

leo-p:
> Should I leave the old `deepo feedback` command available ?

Backward compatibility is always a plus but this means displaying the help will display both functions, which might be confusing.

Also the [documentations](https://docs.deepomatic.com/deepomatic-cli/) needs to be updated when the package will be released.

PR #35, opened 2019-03-04
Support json input

leo-p:
to validate https://airtable.com/tbl0NdAjudtx4qYPZ/viwijmP4oeTteyEJD/reckomCrVky8Yhq0o?blocks=hide

Upgrade deepomatic-cli to support Studio JSON as input.

### DIFF
--- a/deepomatic/cli/io_data.py
+++ b/deepomatic/cli/io_data.py
@@ -392,7 +392,7 @@ class JsonInputData(InputData):
                 elif not 'location' in img:
                     raise NameError(studio_format_error)
                 elif not ImageInputData.is_valid(img['location']):
-                    raise NameError(studio_format_error)
+                    raise NameError('File {} is not valid'.format(img['location']))
         return True
 
     def __init__(self, descriptor, **kwargs):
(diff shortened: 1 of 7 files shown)


deepomatic/cli/io_data.py
```diff
@@ -358,6 +360,84 @@ def get_frame_count(self):
...
+            return False
+
+        # Check if json is a dictionnary
+        try:
+            with open(descriptor) as json_file:
+                json_data = json.load(json_file)
+        except:
+            raise NameError('File {} is not a valid json'.format(descriptor))
+
+        # Check that the json follows the minimum Studio format
```
alemoreau: JSON schema could be used here
https://pypi.org/project/jsonschema/

tests/download_files.py
```diff
@@ -0,0 +1,50 @@
...
+import os
+import json
+import tempfile
+import requests
+
+def download(tmpdir, url, filepath):
+	path = os.path.join(tmpdir, filepath)
+	if not os.path.isdir(os.path.dirname(path)):
+		os.makedirs(os.path.dirname(path))
+	r = requests.get(url)
```
alemoreau: check if file already exists before downloading to speed up tests ? (otherwise it will be downloaded again for each tested command)

leo-p: Initially, it was to make sure all test files were independent, using different temporary directories. Uniting them would mean sharing the temporary directory.

If you think it's worth it, I'll do it.

deepomatic/cli/io_data.py
leo-p: After discussion, will be implemented once the Studio JSON format will be stable.

PR #36, opened 2019-03-06
Migrates output to Studio format

leo-p:
Changes the output JSON format to Deepomatic Studio


deepomatic/cli/cmds/infer.py
```diff
@@ -32,16 +35,24 @@ def run(self):
...
+                name, filename, frame = data
                 if self.workflow is not None and frame is not None:
+                    # Computes prediction and formats them to studio json format
                     prediction = self.workflow.infer(frame).get()
-                    # NOTE: if a higher threshold than self._threshold is set at the recognition version level, we will have less results.
-                    # If we want all of them, we need to set show_discarded=True and filter the one in prediction['discarded'] as well
-                    prediction = [predicted
-                        for predicted in prediction['outputs'][0]['labels']['predicted']
-                        if float(predicted['score']) >= float(self._threshold)]
+                    prediction = transform_json_from_vulcan_to_studio(prediction, name, filename)
```
maingoh: I am not sure we want to convert into studio format all the time. Especially when we don't send to studio. When analyzing a video we might not want this extra computation

deepomatic/cli/cmds/infer.py
```diff
@@ -32,16 +35,24 @@ def run(self):
                     self.workflow.close()
                     return
 
-                name, frame = data
+                name, filename, frame = data
                 if self.workflow is not None and frame is not None:
+                    # Computes prediction and formats them to studio json format
                     prediction = self.workflow.infer(frame).get()
-                    # NOTE: if a higher threshold than self._threshold is set at the recognition version level, we will have less results.
```
maingoh: Why removing this ?

deepomatic/cli/cmds/infer.py
vdel: I agree with @maingoh, we might not want to convert to Studio format all the time. For example when we output OCR results, this is not Studio-compatible (yet ^^). Or when we train a detector and multiple classifiers in the same model: you have more than one output per inference (Thoth will soon be able to train this).

So, it would have been great to kept support for both format, e.g. by adding a new --studio-json output flag. I would also tend to say it is a better idea to keep the --json flag pointing to Vulcan's format as it supports any type of outputs whereas --studio-json would only support single outputs which are label based and is likely to crash in other cases.

BTW, we will obviously make those two format converge over time

vdel: I agree with @maingoh, we might not want to convert to Studio format all the time. For example when we output OCR results, this is not Studio-compatible (yet ^^). Or when we train a detector and multiple classifiers in the same model: you have more than one output per inference (Thoth will soon be able to train this).

So, it would have been great to kept support for both format, e.g. by adding a new --studio-json output flag. I would also tend to say it is a better idea to keep the --json flag pointing to Vulcan's format as it supports any type of outputs whereas --studio-json would only support single outputs which are label based and is likely to crash in other cases.

BTW, we will obviously make those two format converge over time

leo-p: It's easier on the user side if there is only one format to worry about. For instance, this permits performing inference on local images with a fixed threshold and directly uploading all images to vesta. The less JSON manipulation, the better.

As for exotic cases such as OCR and multi-output models, right now, there are none in production to my knowledge.

CC: @thibaut-duguet

deepomatic/cli/cmds/infer.py
leo-p: It's not removed, it's moved [in line 46](https://github.com/Deepomatic/deepocli/pull/36/files#diff-468a395dfeffa63a88b3ee9807677ce8R46).

If a threshold is provided, that's the one we use, otherwise, we use the model threshold (especially useful in the case where classes have different thresholds).

PR #40, opened 2019-03-11
Bulk json output

leo-p:
to validate https://airtable.com/tbl0NdAjudtx4qYPZ/viway1kb3CzTjERth/recCl5bKJAO7FouVi?blocks=hide
- If the JSON output is not a wildcard, store all predictions, not just the last one.
- Adds recursive option for infer/draw/blur input.
- Fixes small bug where discarded results were not accounted for.

### DIFF
--- a/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
+++ b/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
@@ -12,7 +12,9 @@ def transform_json_from_vulcan_to_studio(vulcan_json, name, filename):
 
     # Loop through all images
     img_studio = {'annotated_regions': []}
-    for prediction in vulcan_json['outputs'][0]['labels']['predicted']:
+    all_predictions = vulcan_json['outputs'][0]['labels']['predicted'] + \
+                      vulcan_json['outputs'][0]['labels']['discarded']
+    for prediction in all_predictions:
         # Build studio annotation
         annotation = {
             "tags": [prediction['label_name']],
--- a/deepomatic/cli/io_data.py
+++ b/deepomatic/cli/io_data.py
@@ -726,22 +726,29 @@ class JsonOutputData(OutputData):
     def __init__(self, descriptor, **kwargs):
         super(JsonOutputData, self).__init__(descriptor, **kwargs)
         self._i = 0
+        # Check if the output is a wild card or not
+        try:
+            descriptor % self._i
+            self._all_predictions = None
+        except TypeError:
+            self._all_predictions = {'tags': [], 'images': []}
 
     def __enter__(self):
         self._i = 0
         return self
 
     def __exit__(self, exception_type, exception_value, traceback):
-        pass
+        if self._all_predictions:
+            json_path = os.path.splitext(self._descriptor)[0]
+            save_json_to_file(self._all_predictions, json_path)
 
     def __call__(self, name, frame, prediction):
-        path = self._descriptor
-        try:
-            path = path % self._i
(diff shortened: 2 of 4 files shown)


deepomatic/cli/io_data.py
```diff
@@ -722,22 +726,29 @@ def is_valid(cls, descriptor):
     def __init__(self, descriptor, **kwargs):
         super(JsonOutputData, self).__init__(descriptor, **kwargs)
         self._i = 0
+        # Check if the output is a wild card or not
+        try:
+            descriptor % self._i
```
maingoh: What is `descriptor` ? Is it normal that `_all_predictions` depends on this ?

deepomatic/cli/io_data.py
```diff
@@ -722,22 +726,29 @@ def is_valid(cls, descriptor):
...
+        except TypeError:
+            self._all_predictions = {'tags': [], 'images': []}
 
     def __enter__(self):
         self._i = 0
         return self
 
     def __exit__(self, exception_type, exception_value, traceback):
-        pass
+        if self._all_predictions:
```
maingoh: what if all_predictions is an empty list ? We don't want to store the results ?

deepomatic/cli/io_data.py
```diff
@@ -722,22 +726,29 @@ def is_valid(cls, descriptor):
...
-        except TypeError:
-            pass
-        finally:
-            self._i += 1
-            with open(path, 'w') as file:
-                print_log('Writing %s' % path)
-                json.dump(prediction, file)
+        self._i += 1
+        # If the json is not a wildcard we store prediction to write then to file a the end with the __exit__
+        if self._all_predictions:
```
maingoh: same here ? Are you sure we don't want an `if self._all_predictions is not None` instead ?

deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
```diff
@@ -12,7 +12,9 @@ def transform_json_from_vulcan_to_studio(vulcan_json, name, filename):
 
     # Loop through all images
     img_studio = {'annotated_regions': []}
-    for prediction in vulcan_json['outputs'][0]['labels']['predicted']:
+    all_predictions = vulcan_json['outputs'][0]['labels']['predicted'] + \
```
maingoh: :+1: though you need to specify `show_discarded=True` when infering (it is False by default) here https://github.com/Deepomatic/deepocli/blob/2416b474c42ea26b58a257a14d6dc009a4cb4bcf/deepomatic/cli/workflow/cloud_workflow.py#L44

deepomatic/cli/io_data.py
leo-p: Yes, `descriptor` is the output used in the CLI. It can be for instance `test.json` or `test%02.json`.
Thus we use `_all_predictions` to both store the predictions and indicate if it's a wildcard.

We could also use another variable such as `_is_wildcard` for readability if you think it necessary.

deepomatic/cli/io_data.py
leo-p: `_all_predictions` cannot be an empty list. At the minimum, it's either
- `None`: then we store the result after each inference [see line 752](https://github.com/Deepomatic/deepocli/pull/40/files/d158e733a990e35f6e5228da6c033c2a2b2f61cf#diff-a5f0dd0e9a78c7c2e70327980fe10b36R752).
- `{'tags': [], 'images': []}`: that's all we save to disk.

PR #41, opened 2019-03-12
Output video codecs and formats

leo-p:


deepomatic/cli/io_data.py
```diff
@@ -502,11 +502,7 @@ def is_valid(cls, descriptor):
 
     def __init__(self, descriptor, **kwargs):
         super(VideoOutputData, self).__init__(descriptor, **kwargs)
-        _, ext = os.path.splitext(descriptor)
-        if ext is 'mjpg':
-            fourcc = cv2.VideoWriter_fourcc('M', 'J', 'P', 'G')
-        else:
-            fourcc = cv2.VideoWriter_fourcc('X', '2', '6', '4')
+        fourcc = cv2.VideoWriter_fourcc('M', 'J', 'P', 'G')
```
maingoh: Saving videos in mjpeg is not efficient .. and I don't think it is standard to encapsulate mjpeg into mp4.
Why not installing the good codecs with deepocli ? or detect the one availables ?

PR #42, opened 2019-03-16
Performance improvement

leo-p:
The following PR:
- Separates sending images to the worker and retrieving results in two different threads.
- Takes into account the desired fps when dealing with video to send the minimum number of frames.
- Unifies output method to logging with tqdm.

### DIFF
--- a/deepomatic/cli/cli_parser.py
+++ b/deepomatic/cli/cli_parser.py
@@ -46,9 +46,9 @@ def argparser_init():
         parser.add_argument('-u', '--amqp_url', help="AMQP url for on-premises deployments.")
         parser.add_argument('-k', '--routing_key', help="Recognition routing key for on-premises deployments.")
         parser.add_argument('-t', '--threshold', type=float, help="Threshold above which a prediction is considered valid.", default=None)
+        parser.add_argument('--fps', type=int, help="Video frame rate if applicable.")
 
     for parser in [draw_parser, blur_parser]:
-        parser.add_argument('--fps', type=int, help="Video frame rate if applicable.")
         parser.add_argument('--fullscreen', help="Fullscreen if window output.", action="store_true")
 
     draw_parser.add_argument('--draw_scores', help="Overlays the prediction scores.", action="store_true")
--- a/deepomatic/cli/cli_parser.py
+++ b/deepomatic/cli/cli_parser.py
@@ -48,7 +48,7 @@ def argparser_init():
         parser.add_argument('-t', '--threshold', type=float, help="Threshold above which a prediction is considered valid.", default=None)
 
     for parser in [draw_parser, blur_parser]:
-        parser.add_argument('--output_fps', help="In case of video output, video frame rate.", default=25)
+        parser.add_argument('--fps', type=int, help="Video frame rate if applicable.")
         parser.add_argument('--fullscreen', help="Fullscreen if window output.", action="store_true")
 
     draw_parser.add_argument('--draw_scores', help="Overlays the prediction scores.", action="store_true")
--- a/deepomatic/cli/cmds/infer.py
+++ b/deepomatic/cli/cmds/infer.py
@@ -15,50 +15,44 @@ from deepomatic.cli import io_data
 from deepomatic.cli.workflow import get_workflow
 
 class InferenceThread(threading.Thread):
-    def __init__(self, input_queue, output_queue, **kwargs):
+    def __init__(self, worker_queue, output_queue, workflow, **kwargs):
         threading.Thread.__init__(self, args=(), kwargs=None)
-        self.input_queue = input_queue
+        self.worker_queue = worker_queue
         self.output_queue = output_queue
-        self.daemon = True
-        self.workflow = get_workflow(kwargs)
+        self.workflow = workflow
         self.args = kwargs
-        self._threshold = kwargs.get('threshold', None)
+        self.threshold = kwargs.get('threshold', None)
 
     def run(self):
         try:
             while True:
-                data = self.input_queue.get()
+                data = self.worker_queue.get()
 
                 if data is None:
-                    self.input_queue.task_done()
-                    self.output_queue.put(None)
+                    self.worker_queue.task_done()
+                    self.output_queue.put(None)  # Signals the end of input to the next thread
                     self.workflow.close()
                     return
 
                 result = None
-                name, filename, frame = data
-                if frame is not None:
-                    if self.workflow is not None:
-                        # Computes prediction and formats them to studio json format
-                        prediction = self.workflow.infer(frame).get()
-                        prediction = transform_json_from_vulcan_to_studio(prediction, name, filename)
+                name, filename, frame, inference = data
+                if inference is not None:
+                    prediction = inference.get_predictions()
+                    prediction = transform_json_from_vulcan_to_studio(prediction, name, filename)
 
-                        # Keep only predictions higher than threshold if one is set, otherwise use the model thresholds
-                        kept_pred = []
-                        for pred in prediction['images'][0]['annotated_regions']:
-                            if self._threshold:
-                                if pred['score'] >= self._threshold:
-                                    kept_pred.append(pred)
-                            else:
-                                if pred['score'] >= pred['threshold']:
-                                    kept_pred.append(pred)
-                        prediction['images'][0]['annotated_regions'] = copy.deepcopy(kept_pred)
-                    else:
-                        prediction = {'tags': [], 'images': []}
-
+                    # Keep only predictions higher than threshold if one is set, otherwise use the model thresholds
+                    kept_pred = []
+                    for pred in prediction['images'][0]['annotated_regions']:
+                        if self.threshold:
+                            if pred['score'] >= self.threshold:
+                                kept_pred.append(pred)
+                        else:
+                            if pred['score'] >= pred['threshold']:
+                                kept_pred.append(pred)
+                    prediction['images'][0]['annotated_regions'] = copy.deepcopy(kept_pred)
                     result = self.processing(name, frame, prediction)
 
-                self.input_queue.task_done()
+                self.worker_queue.task_done()
                 self.output_queue.put(result)
         except KeyboardInterrupt:
             logging.info('Stopping output')
--- a/deepomatic/cli/cmds/infer.py
+++ b/deepomatic/cli/cmds/infer.py
@@ -16,7 +16,7 @@ from deepomatic.cli.cmds.studio_helpers.vulcan2studio import transform_json_from
 
 class InferenceThread(threading.Thread):
     def __init__(self, worker_queue, output_queue, workflow, workflow_lock, **kwargs):
-        threading.Thread.__init__(self, args=(), kwargs=None)
+        threading.Thread.__init__(self, args=(), name='WorkerThread', kwargs=None)
         self._worker_queue = worker_queue
         self._output_queue = output_queue
         self._workflow = workflow
--- a/deepomatic/cli/cmds/infer.py
+++ b/deepomatic/cli/cmds/infer.py
@@ -2,7 +2,6 @@ import os
 import sys
 import json
 import copy
-import time
 import logging
 import threading
 
--- a/deepomatic/cli/cmds/infer.py
+++ b/deepomatic/cli/cmds/infer.py
@@ -10,16 +10,16 @@ try:
 except ImportError:
     from queue import Empty
 
-from deepomatic.cli.cmds.studio_helpers.vulcan2studio import transform_json_from_vulcan_to_studio
 from deepomatic.cli import io_data
-from deepomatic.cli.workflow import get_workflow
+from deepomatic.cli.cmds.studio_helpers.vulcan2studio import transform_json_from_vulcan_to_studio
 
 class InferenceThread(threading.Thread):
-    def __init__(self, worker_queue, output_queue, workflow, **kwargs):
+    def __init__(self, worker_queue, output_queue, workflow, workflow_lock, **kwargs):
         threading.Thread.__init__(self, args=(), kwargs=None)
         self.worker_queue = worker_queue
         self.output_queue = output_queue
         self.workflow = workflow
+        self.workflow_lock = workflow_lock
         self.args = kwargs
         self.threshold = kwargs.get('threshold', None)
 
@@ -37,7 +37,8 @@ class InferenceThread(threading.Thread):
                 result = None
                 name, filename, frame, inference = data
                 if inference is not None:
-                    prediction = inference.get_predictions()
+                    with self.workflow_lock:
+                        prediction = inference.get_predictions()
                     prediction = transform_json_from_vulcan_to_studio(prediction, name, filename)
 
                     # Keep only predictions higher than threshold if one is set, otherwise use the model thresholds
--- a/deepomatic/cli/cmds/infer.py
+++ b/deepomatic/cli/cmds/infer.py
@@ -29,7 +29,6 @@ class InferenceThread(threading.Thread):
         try:
             while 1:
                 data = self.worker_queue.get()
-
                 if data is TERMINATION_MSG:
                     self.worker_queue.task_done()
                     self.output_queue.put(TERMINATION_MSG)
@@ -45,8 +44,8 @@ class InferenceThread(threading.Thread):
                         self.workflow.close()
                         return
                 else:
-                    result = None
-                    name, filename, frame, inference = data
+                    frame_processed = None
+                    frame_number, name, filename, frame, inference = data
                     if inference is not None:
                         with self.workflow_lock:
                             prediction = inference.get_predictions()
@@ -67,10 +66,10 @@ class InferenceThread(threading.Thread):
                                 if pred['score'] >= pred['threshold']:
                                     kept_pred.append(pred)
                         prediction['images'][0]['annotated_regions'] = copy.deepcopy(kept_pred)
-                        result = self.processing(name, frame, prediction)
+                        frame_processed = self.processing(name, frame, prediction)
 
                     self.worker_queue.task_done()
-                    self.output_queue.put(result)
+                    self.output_queue.put((frame_number, frame_processed))
         except KeyboardInterrupt:
             logging.info('Stopping output')
             while not self.output_queue.empty():
--- a/deepomatic/cli/cmds/infer.py
+++ b/deepomatic/cli/cmds/infer.py
@@ -17,49 +17,49 @@ from deepomatic.cli.cmds.studio_helpers.vulcan2studio import transform_json_from
 class InferenceThread(threading.Thread):
     def __init__(self, worker_queue, output_queue, workflow, workflow_lock, **kwargs):
         threading.Thread.__init__(self, args=(), kwargs=None)
-        self.worker_queue = worker_queue
-        self.output_queue = output_queue
-        self.workflow = workflow
-        self.workflow_lock = workflow_lock
-        self.args = kwargs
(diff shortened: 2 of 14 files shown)


deepomatic/cli/cmds/infer.py
```diff
@@ -2,64 +2,74 @@
...
+from deepomatic.cli.cmds.studio_helpers.vulcan2studio import transform_json_from_vulcan_to_studio
 
 class InferenceThread(threading.Thread):
-    def __init__(self, input_queue, output_queue, **kwargs):
+    def __init__(self, worker_queue, output_queue, workflow, workflow_lock, **kwargs):
         threading.Thread.__init__(self, args=(), kwargs=None)
-        self.input_queue = input_queue
+        self.worker_queue = worker_queue
         self.output_queue = output_queue
-        self.daemon = True
```
vdel: Why removing this line ? It is important so that CTRL+C effectively kills the whole process, deepocli will hang forever on CTRL+C

deepomatic/cli/cmds/infer.py
```diff
@@ -2,64 +2,74 @@
...
 
     def run(self):
         try:
-            while True:
-                data = self.input_queue.get()
-
-                if data is None:
-                    self.input_queue.task_done()
-                    self.output_queue.put(None)
+            while 1:
```
vdel: Use `While True` please

deepomatic/cli/cmds/infer.py
```diff
@@ -2,64 +2,74 @@
...
+                        self.workflow.close()
+                        return
+                else:
+                    frame_processed = None
+                    frame_number, name, filename, frame, inference = data
+                    if inference is not None:
+                        with self.workflow_lock:
+                            prediction = inference.get_predictions()
+                            # If the prediction is not finished, then put the data back in the queue
+                            if prediction is None:
```
vdel: Why do you put pack in the queue ? Frames are (or at least 'can be') processed in sequential order so it's okay to block on this one until it's done

deepomatic/cli/cmds/infer.py
```diff
@@ -2,64 +2,74 @@
...
-                        # Computes prediction and formats them to studio json format
-                        prediction = self.workflow.infer(frame).get()
                         prediction = transform_json_from_vulcan_to_studio(prediction, name, filename)
-
                         # Keep only predictions higher than threshold if one is set, otherwise use the model thresholds
                         kept_pred = []
                         for pred in prediction['images'][0]['annotated_regions']:
-                            if self._threshold:
-                                if pred['score'] >= self._threshold:
+                            if self.threshold:
```
vdel: Do not make it public. If you need to access it, make a @property accessor

deepomatic/cli/io_data.py
```diff
@@ -7,13 +7,16 @@
...
 
 try:
     from Queue import Queue, LifoQueue, Empty
 except ImportError:
     from queue import Queue, LifoQueue, Empty
 
-INPUT_QUEUE_MAX_SIZE = 50
+QUEUE_MAX_SIZE = 50
+END_OF_STREAM_MSG = "__END_OF_STREAM__"
+TERMINATION_MSG = "__TERMINATION_MSG__"
```
vdel: Seems like a big hack. What do you need those for ?

deepomatic/cli/io_data.py
```diff
@@ -83,14 +86,31 @@ def input_loop(kwargs, worker_thread):
 
     # For realtime, queue should be LIFO
     input_queue = LifoQueue() if inputs.is_infinite() else Queue()
+    worker_queue = LifoQueue() if inputs.is_infinite() else Queue()
```
vdel: Using a LifoQueue is a bad idea here. If I understand well, you use it to skip frames and get the most recent ones ? What about the case where frames will pile-up faster than what you can pop ? It will likely lead to out of memory after some time

deepomatic/cli/io_data.py
```diff
@@ -83,14 +86,31 @@ def input_loop(kwargs, worker_thread):
...
     output_queue = LifoQueue() if inputs.is_infinite() else Queue()
 
-    worker = worker_thread(input_queue, output_queue, **kwargs)
-    worker.start()
+    # Initialize workflow for mutual use between input and worker threads:
+    #   - input thread uses it to send frames
+    #   - worker thread uses it to retrieve predictions
+    # Note: the workflow is closed by the worker thread
+    workflow = get_workflow(kwargs)
+    workflow_lock = threading.Lock()  # Controls access to underlying socket
```
vdel: Which socket are you referring to ?

deepomatic/cli/io_data.py
```diff
@@ -126,35 +143,92 @@ def input_loop(kwargs, worker_thread):
...
+        threading.Thread.__init__(self, args=(), kwargs=None)
+        self.input_queue = input_queue
+        self.worker_queue = worker_queue
+        self.workflow = workflow
+        self.workflow_lock = workflow_lock
+        self.args = kwargs
+
+    def run(self):
+        try:
+            while 1:
```
vdel: Use `While True` please. Readability and Understandability is more important than saving typing 3 characters. `while 1` looks like a potential mistake.

deepomatic/cli/io_data.py
leo-p: I've added a max size to avoir out of memories but I think we need to rethink the whole infinite stream pipeline.
Might be a good idea to tackle when migrating to Async IO.

deepomatic/cli/io_data.py
```diff
@@ -73,24 +72,45 @@ def get_output(descriptor, kwargs):
...
+    pbar = tqdm(total=max_value, file=tqdmout, desc='Input processing', smoothing=0)
 
     # For realtime, queue should be LIFO
-    input_queue = LifoQueue() if inputs.is_infinite() else Queue()
-    output_queue = LifoQueue() if inputs.is_infinite() else Queue()
-
-    worker = worker_thread(input_queue, output_queue, **kwargs)
-    worker.start()
+    # TODO: might need to rethink the whole pipeling for infinite streams
+    input_queue = LifoQueue(maxsize=QUEUE_MAX_SIZE) if inputs.is_infinite() else Queue()
```
maingoh: maxsize only for `LifoQueue` ?

deepomatic/cli/io_data.py
```diff
@@ -101,21 +121,18 @@ def input_loop(kwargs, worker_thread):
                     except Empty:
                         break
 
-            while input_queue.qsize() > INPUT_QUEUE_MAX_SIZE:
+            while input_queue.qsize() > QUEUE_MAX_SIZE or worker_queue.qsize() > QUEUE_MAX_SIZE:
                 time.sleep(1)
 
-            input_queue.put(frame)
+            data = (frame_number,) + frame  # Python 2.7 compatiblity
```
maingoh: What was the problem in python 2 ?

deepomatic/cli/io_data.py
```diff
@@ -126,35 +143,92 @@ def input_loop(kwargs, worker_thread):
...
+        self.worker_queue = worker_queue
+        self.workflow = workflow
+        self.workflow_lock = workflow_lock
+        self.args = kwargs
+
+    def run(self):
+        try:
+            while True:
+                data = self.input_queue.get()
+                if data is END_OF_STREAM_MSG or data is TERMINATION_MSG:
```
maingoh: we should be able to have a ThreadBase that handle all of this in a factorized way

deepomatic/cli/io_data.py
```diff
@@ -126,35 +143,92 @@ def input_loop(kwargs, worker_thread):
...
+                frame_number, name, filename, frame = data
+                if frame is not None:
+                    if self.workflow is not None:
+                        with self.workflow_lock:
+                            inference = self.workflow.infer(frame)
+
+                self.input_queue.task_done()
+                self.worker_queue.put((frame_number, name, filename, frame, inference))
+        except KeyboardInterrupt:
+            logging.info('Stopping output')
```
maingoh: Same here

deepomatic/cli/io_data.py
```diff
@@ -126,35 +143,92 @@ def input_loop(kwargs, worker_thread):
...
-                    self.queue.task_done()
+                    elif data is END_OF_STREAM_MSG:
+                        self.output_queue.task_done()
+                        if not self.output_queue.empty():
+                            self.output_queue.put(END_OF_STREAM_MSG)
+                            continue
+                        else:
+                            return
+                    else:
+                        queue_frame_number, queue_frame_processed = data
```
maingoh: data is never the same type .. not easy for readability and might lead to bugs

deepomatic/cli/io_data.py
```diff
@@ -240,10 +314,20 @@ def is_valid(cls, descriptor):
 
     def __init__(self, descriptor, **kwargs):
         super(VideoInputData, self).__init__(descriptor, **kwargs)
-        self._cap = None
         self._i = 0
         self._name = '%s_%s_%s' % (self._name, '%05d', self._reco)
         self._cap = cv2.VideoCapture(self._descriptor)
+        if self._cap is not None:
+            raw_fps = self._cap.get(cv2.CAP_PROP_FPS)
+            desired_fps = min(kwargs['fps'], raw_fps) if kwargs['fps'] else raw_fps
```
maingoh: You should log when doing magic (changing the fps that the user asked for)

deepomatic/cli/io_data.py
```diff
@@ -512,7 +592,7 @@ def __init__(self, descriptor, **kwargs):
         elif ext == '.mp4':
             fourcc = cv2.VideoWriter_fourcc('m', 'p', '4', 'v')
         self._fourcc = fourcc
-        self._fps = kwargs.get('output_fps', 25)
+        self._fps = kwargs['fps'] if kwargs['fps'] else 25
```
maingoh: nit: `self._fps = kwargs['fps'] or DEFAULT_FPS`

deepomatic/cli/io_data.py
```diff
@@ -672,16 +749,16 @@ def __call__(self, name, frame, prediction):
...
         return self
+
     def __exit__(self, exception_type, exception_value, traceback):
         pass
 
 class DisplayOutputData(OutputData):
     def __init__(self, **kwargs):
         super(DisplayOutputData, self).__init__(None, **kwargs)
-        self._fps = kwargs.get('output_fps', 25)
+        self._fps = kwargs.get('fps', 25)
```
maingoh: Use a global variable for the default fps

deepomatic/cli/workflow/rpc_workflow.py
```diff
@@ -2,19 +2,16 @@
...
-except ImportError:
-    can_use_rpc = False
-
-if can_use_rpc:
     from deepomatic.rpc import v07_ImageInput, BINARY_IMAGE_PREFIX
+    from deepomatic.rpc.response import wait_responses
     from deepomatic.rpc.helpers.v07_proto import create_recognition_command_mix
     from deepomatic.rpc.helpers.proto import create_v07_images_command
     from google.protobuf.json_format import MessageToDict
+    RPC_PACKAGES_USABLE = True
```
maingoh: Please put back as before: 
- first we want to check that the rpc module is importable
- if imported successfully the rest should import as well, otherwise it is probably a version problem in which case we shouldn't silent the import error
If only one function failed to import we will not know which one and why
You can add a comment explaining this (to avoid a newer developer to do this again :) )

deepomatic/cli/workflow/rpc_workflow.py
```diff
@@ -25,14 +22,16 @@ def __init__(self, correlation_id, consumer):
...
-            response = self._consumer.get(self._correlation_id, timeout=60)
-            outputs = response.to_parsed_result_buffer()
-            return {
-                'outputs': [{
-                    'labels': MessageToDict(output.labels, including_default_value_fields=True, preserving_proto_field_name=True)
-                } for output in outputs]
-            }
+        def get_predictions(self):
+            # TODO: Enforce a certain level of force waiting for live streams
+            response_done, response_pending = wait_responses(self._consumer, [self._correlation_id], timeout=0.010)
```
maingoh: Why did you change this ? You only wait for one response here, you didn't know what exception to catch ?

deepomatic/cli/cli_parser.py
```diff
@@ -46,9 +46,9 @@ def argparser_init():
         parser.add_argument('-u', '--amqp_url', help="AMQP url for on-premises deployments.")
         parser.add_argument('-k', '--routing_key', help="Recognition routing key for on-premises deployments.")
         parser.add_argument('-t', '--threshold', type=float, help="Threshold above which a prediction is considered valid.", default=None)
+        parser.add_argument('--fps', type=int, help="Video frame rate if applicable.")
```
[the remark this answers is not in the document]
leo-p: It's actually the fps of the whole video pipeline, input and output.
The idea is not to send frames that are going to be ignored later on, so we send only the frames that will be used in the output.

deepomatic/cli/cmds/infer.py
```diff
@@ -2,47 +2,59 @@
...
-                    self.input_queue.task_done()
-                    self.output_queue.put(None)
-                    self.workflow.close()
+                data = self._worker_queue.get()
+                if data is TERMINATION_MSG:
+                    self._worker_queue.task_done()
+                    self._output_queue.put(TERMINATION_MSG)
+                    self._workflow.close()
                     return
+                elif data is END_OF_STREAM_MSG:
```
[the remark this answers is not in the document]
leo-p: Duly noted 👍

deepomatic/cli/io_data.py
leo-p: Yes, because LifoQueue is used only for infinite input. If the input is not infinite we don't want to lose any frame.

deepomatic/cli/io_data.py
leo-p: There is a new parameter to add to the data list and the star parameter doesn't place nice in Python 2.7.

deepomatic/cli/workflow/rpc_workflow.py
leo-p: I purposefully wait for one response, the best would even be not to wait with for instance a `timeout=False`. Basically either the prediction for this correlation_id is ready and we manage it, or it's not and we move on.

deepomatic/cli/cmds/infer.py
```diff
@@ -2,47 +2,59 @@
...
         try:
             while True:
-                data = self.input_queue.get()
-
-                if data is None:
-                    self.input_queue.task_done()
-                    self.output_queue.put(None)
-                    self.workflow.close()
+                data = self._worker_queue.get()
+                if data is TERMINATION_MSG:
```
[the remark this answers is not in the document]
leo-p: FTR: will be done in another PR

deepomatic/cli/cmds/infer.py
```diff
@@ -2,47 +2,59 @@
...
+                        continue
+                    else:
+                        self._output_queue.put(END_OF_STREAM_MSG)
+                        self._workflow.close()
+                        return
+                else:
+                    frame_processed = None
+                    frame_number, name, filename, frame, inference = data
+                    if inference is not None:
+                        with self._workflow_lock:
```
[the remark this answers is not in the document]
leo-p: FTR: will be done in another PR.

deepomatic/cli/io_data.py
leo-p: FTR: will be done in another PR.

deepomatic/cli/io_data.py
alemoreau: Setting a maxsize for a queue won't lose any frame, it will simply block the input thread until the worker thread has finished. It should be set to prevent the input thread from filling the memory

alemoreau: I think it should not be a LifoQueue but a custom object using traditional thread synchronization techniques

deepomatic/cli/io_data.py
```diff
@@ -73,24 +73,57 @@ def get_output(descriptor, kwargs):
...
+    # Define threads
+    #   - input: prepares input and send it to worker
+    #   - worker: retrieves prediction from worker
+    #   - output: transforms predictions into outputs
+    input_thread = InputThread(input_queue, worker_queue, workflow, workflow_lock, **kwargs)
+    worker_thread = WorkerThread(worker_queue, output_queue, workflow, workflow_lock, **kwargs)
     output_thread = OutputThread(output_queue, on_progress=lambda i: pbar.update(1), **kwargs)
+
+    # Start threads
+    daemon = True  # Makes all threads daemon threads, must be set before starting threads
```
alemoreau: It should be set as a property for each thread object

deepomatic/cli/io_data.py
```diff
@@ -126,35 +156,92 @@ def input_loop(kwargs, worker_thread):
...
-        with get_output(self.args.get('output', None), self.args) as output:
+        with get_output(self.args.get('output', None), self.args) as output_processing:
             try:
-                while True:
-                    i += 1
-                    data = self.queue.get()
-                    if data is None:
-                        self.queue.task_done()
+                frame_to_process = 0
+                while 1:
```
vdel: Please use `while True:`

deepomatic/cli/io_data.py
```diff
@@ -126,35 +156,92 @@ def input_loop(kwargs, worker_thread):
                 input_queue.task_done()
             except Empty:
                 break
-        input_queue.put(None)
-
-        worker.join()
+        input_queue.put(TERMINATION_MSG)
```
vdel: You do not need all this try catch if you use daemon = True (see my other comment).
I think you can remove all this and get rid of TERMINATION_MSG

deepomatic/cli/io_data.py
```diff
@@ -126,35 +156,92 @@ def input_loop(kwargs, worker_thread):
...
+                inference = None
+                frame_number, name, filename, frame = data
+                if frame is not None:
+                    if self.workflow is not None:
+                        with self.workflow_lock:
+                            inference = self.workflow.infer(frame)
+
+                self.input_queue.task_done()
+                self.worker_queue.put((frame_number, name, filename, frame, inference))
+        except KeyboardInterrupt:
```
vdel: You do not need all this try catch if you use daemon = True (see my other comment).
I think you can remove all this and get rid of TERMINATION_MSG

deepomatic/cli/cmds/infer.py
```diff
@@ -53,23 +65,20 @@ def run(self):
...
-                        prediction = {'tags': [], 'images': []}
-
-                    result = self.processing(name, frame, prediction)
+                        frame_processed = self.processing(name, frame, prediction)
 
-                self.input_queue.task_done()
-                self.output_queue.put(result)
+                    self._worker_queue.task_done()
+                    self._output_queue.put((frame_number, frame_processed))
         except KeyboardInterrupt:
```
vdel: You do not need all this try catch if you use daemon = True (see my other comment).
I think you can remove all this and get rid of TERMINATION_MSG

deepomatic/cli/io_data.py
vdel: (this is not mandatory but it would improve the readability of the whole code)

deepomatic/cli/io_data.py
vdel: (this is not mandatory but it would improve the readability of the whole code)

deepomatic/cli/cmds/infer.py
vdel: (this is not mandatory but it would improve the readability of the whole code)

maingoh:
Closing as this PR has been included in #43

PR #43, opened 2019-04-03
Refactoring for better performances

### LINKED ISSUE #38: None frame when drawing ?
None frame should be received in Infer Thread when the program has to exit. Though it seems some None values are received when the program is not done to process an entire directory ?

maingoh:
Based on #42 

* Separates sending images to the worker and retrieving results in two different threads
* Takes into account the desired fps when dealing with video to send the minimum number of frames.
* Unifies output method to logging with tqdm.
* Split `io_data.py` into `input_data.py` and `output_data.py` + moved draw and blur postprocessing in `infer.py`
* Putting back run format as default and added `-s/--studio-format`
* Fixed KeyboardInterrupt and waiting of threads (all threads inherit from `ThreadBase`)
* Exit in error if one thread fail (test were not failing if threads raised exception)
* Using a uniq `Frame` object that we transfer between queues
* Factorized code whenever possible
* some pep8

Close #38


deepomatic/cli/cmds/infer.py
```diff
@@ -1,82 +1,175 @@
...
-    try:
-        io_data.input_loop(args, InferenceThread)
-    except KeyboardInterrupt:
-        pass
+            except Empty:
+                if not self.frames_to_check_first:
+                    return
+                frame = self.frames_to_check_first.pop(0)
+
+        predictions = frame.inference_async_result.get_predictions()
```
alemoreau: What about using `wait_responses` from deepomatic.rpc for waiting on all `frames_to_check_first` ?

deepomatic/cli/cmds/infer.py
```diff
@@ -1,82 +1,175 @@
...
+    def fill_predictions(self, predictions, new_predicted, new_discarded):
+        for prediction in predictions:
+            if prediction['score'] >= self.threshold:
+                new_predicted.append(prediction)
+            else:
+                new_discarded.append(prediction)
+
+    def loop_impl(self):
+        # we keep an internal cache that we re-check when it is big or that the input_queue is empty
+        if len(self.frames_to_check_first) > 10:
```
alemoreau: 10 -> env variable ? (it should depend on the worker count to maximize throughput)

leo-p: Will this `10` need to be adjusted depending on the machine?

deepomatic/cli/workflow/rpc_workflow.py
```diff
@@ -25,26 +30,30 @@ def __init__(self, correlation_id, consumer):
...
-            }
+        def get_predictions(self):
+            # TODO: Enforce a certain level of force waiting for live streams
+            response = self._consumer.get(self._correlation_id, timeout=None)
+            if response is not None:
+                outputs = response.to_parsed_result_buffer()
+                predictions = {'outputs': [{'labels': MessageToDict(output.labels, including_default_value_fields=True, preserving_proto_field_name=True)} for output in outputs]}
+                return predictions
+            else:
+                time.sleep(0.010)
```
alemoreau: Wouldn't it be more flexible to let the thread calling this method handle the sleep ?

deepomatic/cli/cmds/infer.py
```diff
@@ -1,82 +1,175 @@
...
+                labels['discarded'] = new_discarded
+
+        frame.predictions = predictions
+
+        if self.postprocessing is not None:
+            self.postprocessing(frame)
+        else:
+            frame.image_output = frame.image  # we output the original image
+
+        # self.input_queue.task_done()
```
leo-p: Why is this commented? Debug artifact?

deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
```diff
@@ -12,8 +7,7 @@ def transform_json_from_vulcan_to_studio(vulcan_json, name, filename):
 
     # Loop through all images
     img_studio = {'annotated_regions': []}
-    all_predictions = vulcan_json['outputs'][0]['labels']['predicted'] + \
-                      vulcan_json['outputs'][0]['labels']['discarded']
+    all_predictions = vulcan_json['outputs'][0]['labels']['predicted']
```
leo-p: Do we want to remove discarded predictions? Do we use this function before applying a subsequent threshold that might be lower?

deepomatic/cli/workflow/rpc_workflow.py
maingoh: Yes, I wanted to remove this, I saw it this week end, this should be done in the thread and probably shouldn't wait that much

deepomatic/cli/cmds/infer.py
maingoh: We should benchmark but since currently there is a postprocessing in the same thread, this is probably not optimal. We want to keep doing things whenever possible and probably not wait for the whole batch.
I made the wait_responses function so that it is easy for the user to wait on multiple responses but here I think it will be slower

deepomatic/cli/cmds/studio_helpers/task.py
```diff
@@ -9,7 +10,7 @@ def retrieve(self, task_id, wait=True):
         ret = self._helper.get('manage/tasks/{}/'.format(task_id))
         if not wait:
             return ret
-        while ret['next'] != None and ret['status'] != 'SUCCESS':
+        while ret['next'] is not None and ret['status'] != 'SUCCESS':
             if ret['status'] in ('FAILURE', 'REVOKED'):
                 raise RuntimeError("Task {} stoped with status".format(task_id))
```
alemoreau:
```suggestion
                raise RuntimeError("Task {} stopped with status".format(task_id))
```

deepomatic/cli/cmds/infer.py
maingoh: What we can probably do is moving the postprocessing in another thread and then use `wait_responses`

alemoreau: Alright, so `wait_responses` is not meant to be used with `timeout=None` ?

deepomatic/cli/cmds/infer.py
maingoh: It might or might be automatically calculated. Though I don't think it will change much (I don't have enough information to know if it needs to be changed/dynamic

deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
maingoh: Now `vulcan_json['outputs'][0]['labels']['predicted']` is updated depending on the provided `threshold` before going here. Before it was filtered after.

deepomatic/cli/cmds/infer.py
maingoh: I will make it an env vars, though this shouldn't impact much the performances, and when I have more benchmark, probably this should be dynamic or stay hardcoded

tests/test_blur.py
```diff
@@ -18,103 +18,135 @@
 
 # ------- Image Input Tests ------------------------------------------------------------------------------------------ #
 
+
```
vdel: FYI, this looks like a very good use case for pytest parametrize (example here: https://github.com/Deepomatic/thoth/blob/master/e2e/test_oxford_pets.py#L19)

maingoh: Indeed, added this to https://github.com/Deepomatic/deepocli/issues/45

PR #44, opened 2019-04-04
Multiple outputs

alemoreau:
I needed to output raw frames (not drawn) and associated predictions, so I made this.
You can now use multiple outputs:
deepo infer -i 192.168.11.21_2019-02-28_12-39-29.mp4 **-o frames_%05d.jpg predictions_%05d.json** -k worker_queue-u amqp://guest:password@0.0.0.0:5672/ -r 30819

Screenshot from 2019-04-04 11-57-23

I also removed --json since I think it is redundant and less flexible than this.

### DIFF
--- a/tests/test_blur.py
+++ b/tests/test_blur.py
@@ -13,6 +13,7 @@ image_output = os.path.join(dir_output, 'image_output%4d.jpg')
 video_output = os.path.join(dir_output, 'video_output%4d.mp4')
 json_output = os.path.join(dir_output, 'test_output%4d.json')
 directory_output = dir_output
+outputs = [std_output, image_output, video_output, json_output, directory_output]
 
 
 # ------- Image Input Tests ------------------------------------------------------------------------------------------ #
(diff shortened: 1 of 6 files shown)


tests/test_blur.py
```diff
@@ -13,6 +13,7 @@
 video_output = os.path.join(dir_output, 'video_output%4d.mp4')
 json_output = os.path.join(dir_output, 'test_output%4d.json')
 directory_output = dir_output
+outputs = [std_output, image_output, video_output, json_output, directory_input]
```
leo-p:
```suggestion
outputs = [std_output, image_output, video_output, json_output, directory_output]
```

tests/test_draw.py
```diff
@@ -13,7 +13,7 @@
 video_output = os.path.join(dir_output, 'video_output%4d.mp4')
 json_output = os.path.join(dir_output, 'test_output%4d.json')
 directory_output = dir_output
-
+outputs = [std_output, image_output, video_output, json_output, directory_input]
```
leo-p:
```suggestion
outputs = [std_output, image_output, video_output, json_output, directory_output]
```

tests/test_infer.py
```diff
@@ -7,14 +7,23 @@
 
 # Input to test: Image, Video, Directory, Json
 image_input, video_input, directory_input, json_input, dir_output = init_files_setup()
-# Output to test: Stdout, Json, Directory
+# Output to test: Image, Video, Stdout, Json, Directory
```
leo-p: Does it make sense to output image and videos for inference?

alemoreau: I needed to output raw frames (not drawn) and associated predictions for a video.
I think it makes even more sense for stream input (where the frame associated to the output prediction would be lost, or hard to associate to the correct prediction)

leo-p: It does make a lot of sense 👍! Also if you're inferring on a video and using a different fps than the one in the video.

PR #49, opened 2019-04-10
Harmonizes thread implementation

leo-p:
- Fixes bug where `studio add_images` threads wouldn't stop.
- Harmonizes thread implementation using `thread_base.py`
- Removes useless parameter for `on_progress` function

### DIFF
--- a/deepomatic/cli/cmds/studio_helpers/file.py
+++ b/deepomatic/cli/cmds/studio_helpers/file.py
@@ -12,7 +12,7 @@ from .task import Task
 from ...common import TqdmToLogger
 from ... import thread_base
 if sys.version_info >= (3, 0):
-    import queue as Queue
+    import queue as Queue, Empty
 else:
     import Queue
 
@@ -26,12 +26,11 @@ class UploadImageThread(thread_base.ThreadBase):
         self.args = kwargs
         self.on_progress = on_progress
         self._helper = helper
-        self._input_queue = input_queue
         self._task = task
 
     def loop_impl(self):
         try:
-            url, data, file = self._input_queue.get(timeout=thread_base.POP_TIMEOUT)
+            url, data, file = self.input_queue.get(timeout=thread_base.POP_TIMEOUT)
         except Empty:
             return
 
@@ -40,9 +39,10 @@ class UploadImageThread(thread_base.ThreadBase):
                 rq = self._helper.post(url, data={"meta": data}, content_type='multipart/form', files={"file": fd})
                 self._task.retrieve(rq['task_id'])
         except RuntimeError as e:
-            logging.error(e, file=sys.stderr)
+            logging.error(e)
+            logging.error("URL {} with file {} failed".format(url, file))
 
-        self._input_queue.task_done()
+        self.input_queue.task_done()
         if self.on_progress:
             self.on_progress()
 
@@ -53,7 +53,7 @@ class File(object):
         if not task:
             task = Task(helper)
         self._task = task
-        self._input_queue = Queue.Queue()
+        self.input_queue = Queue.Queue()
         self.total_files = 0
 
 
@@ -62,7 +62,7 @@ class File(object):
             # If it's an file, add it to the queue
             if file.split('.')[-1].lower() != 'json':
                 tmp_name = uuid.uuid4().hex
-                self._input_queue.put((
+                self.input_queue.put((
                     'v1-beta/datasets/{}/commits/{}/images/'.format(dataset_name, commit_pk),
                     json.dumps({'location': tmp_name}),
                     file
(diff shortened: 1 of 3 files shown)


deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -10,48 +10,41 @@
...
-    while not q.empty():
-        q.get()
-        q.task_done()
+class UploadImageThread(thread_base.ThreadBase):
+    def __init__(self, exit_event, input_queue, helper, task, on_progress=None, **kwargs):
+        super(UploadImageThread, self).__init__(exit_event, 'UploadImageThread', input_queue)
+        self.args = kwargs
+        self.on_progress = on_progress
+        self._helper = helper
+        self._input_queue = input_queue
```
maingoh: the _input_queue is already set by the ThreadBase

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -10,48 +10,41 @@
...
-                logging.error('Annotation format for file named {} is incorrect'.format(file), file=sys.stderr)
-            pbar.update(1)
-            q.task_done()
-            lock.acquire()
-            count += 1
-            lock.release()
-        except Queue.Empty:
-            pass
+        except RuntimeError as e:
+            logging.error(e, file=sys.stderr)
```
maingoh: logging is always done in stderr, also we should log which url failed

deepomatic/cli/input_data.py
```diff
@@ -83,7 +83,7 @@ def input_loop(kwargs, postprocessing=None):
     exit_event = threading.Event()
     send_inference_thread = SendInferenceThread(exit_event, input_queue, worker_queue, workflow, **kwargs)
     result_inference_thread = ResultInferenceThread(exit_event, worker_queue, output_queue, workflow, postprocessing=postprocessing, **kwargs)
-    output_thread = OutputThread(exit_event, output_queue, on_progress=lambda i: pbar.update(1), **kwargs)
```
maingoh: it still works without parameter ?

deepomatic/cli/output_data.py
```diff
@@ -87,7 +87,7 @@ def loop_impl(self):
             output.output_frame(frame)
         self.frame_to_output += 1
         if self.on_progress:
-            self.on_progress(self.frame_to_output)
```
maingoh: why removing this ?

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -111,38 +102,70 @@ def post_files(self, dataset_name, files, org_slug, is_json=False):
...
-        if count == total_files:
-            logging.info("All {} files have been uploaded.".format(count))
-        else:
-            logging.info("{} files out of {} have been uploaded.".format(count, total_files))
+
+        # If the process encountered an error, the exit code is 1.
+        # If the process is interrupted using SIGINT (ctrl + C) or SIGTERM, the threads are stopped, and
+        # the exit code is 0.
+        if exit_event.is_set():
+            sys.exit(1)
```
maingoh: no exit code 2 when double KeyboardInterrupt ?

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -111,38 +102,70 @@ def post_files(self, dataset_name, files, org_slug, is_json=False):
...
-        while not q.empty() and run:
-            time.sleep(2)
-        run = False
+        # Start threads
+        for t in threads:
+            t.start()
 
-        # Close properly
-        q.join()
+        while True:
```
maingoh: This is sad that it is not integrated with the rest of the code, this would avoid duplication of code. I will open an issue for this

deepomatic/cli/input_data.py
leo-p: Yes!

deepomatic/cli/output_data.py
leo-p: Because the parameter was useless, we always perform `pbar.update(1)` which increments the bar by 1.

deepomatic/cli/cmds/studio_helpers/file.py
leo-p: I'm not sure about his. Pressing KeyboardInterrupt twice or more times doesn't actually change anything, compared to `infer` where it stops all threads. Here pressing it once or thrice results in the same stopping mechanism.

Thus I went with the 1-KeyboardInterrupt-press exit code.

maingoh: Alright, let see this when it will be properly integrated with the rest

leo-p:
Tested the upload on 2000 images and got 2000 images on Studio.

Issue #19, opened 2018-12-20
Debug mode


Alexis-Jacob:
It would be nice if we could activate a debug mode such as 
```python
if os.getenv('debug'):
    import requests
    import logging
    logging.basicConfig(level=logging.DEBUG)
```

thomas-riccardi:
maybe call that `verbose`.

maingoh:
Also we should:
- use `logger = logging.getLogger(__name__)`
- call `logging.basicConfig()` only once in the init of the package

To allow for debug or any other log level I think it is better to use `logging.basicConfig(level=os.getenv('LOG_LEVEL', 'INFO'))` like in `deepomatic-api`

maingoh:
So here we just need to choose what we want to log in debug and what we want to log in info. The rest has been done. @Alexis-Jacob any memory of what you wanted to ignore ?

Alexis-Jacob:
No, my purpose was to see the requests done by deepocli a runtime, but nothing to ignore is occuring to my mind

maingoh:
Alright, we can also put this kind of log in debug https://github.com/Deepomatic/deepocli/blob/103ba48991ff446a0235f0ddb33c08957f001c1f/deepomatic/cli/output_data.py#L26

PR #51, opened 2019-04-10
Gevent optimization + Regulate infer requests + factorized tests

### LINKED ISSUE #45: Improve tests
We should check that outputs works well (all files exists on the disk and are well formated)

### LINKED ISSUE #46: Regulate the number of requests pushed
Since #43 we now push requests with a lot of advance which make the requests accumulating in rabbitmq. In order to avoid disk/memory leak on rabbitmq for long streams we should regulate the number of message in the queue by synchronizing `SendInferenceThread` and `ResultInferenceThread`

maingoh:
Close #45
Close #46 
- Added maxsize to queues for regulation (avoid to push too many requests to rabbitmq at once)
- Using gevent for I/O optimization and pools of executors to maximize throughput (reaching 60 fps on a classifier with rpc, the worker is currently the bottleneck)
- Better handling of shutdown and signals
- Factorized tests + using pytest parametrized + checking files dumped on the disk
- Handling timeouts and error during inference
- Centralized logging (partly addressing #19)


tests/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    with open(json_pth, 'r') as json_file:
+        json_data = json.load(json_file)
+    json_data['images'][0]['location'] = single_img_pth
+    with open(json_pth, 'w') as json_file:
+        json.dump(json_data, json_file)
+
+    return single_img_pth, video_pth, img_dir_pth, json_pth
+
+
+def run_cmd(cmds, inp, outputs, *args, **kwargs):
```
leo-p:
```suggestion
def run_cmd(cmds, input, outputs, *args, **kwargs):
```

tests/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    extra_opts = kwargs.pop('extra_opts', [])
+    absolute_outputs = []
+    with create_tmp_dir() as tmpdir:
+        for output in outputs:
+            if output == STD_OUTPUT:
+                absolute_outputs.append(output)
+            else:
```
leo-p: What about `WINDOW_OUTPUT` ?

tests/utils.py
```diff
@@ -0,0 +1,117 @@
...
+from contextlib import contextmanager
+from deepomatic.cli.cli_parser import run
+
+
+# Output to test: Image, Video, Stdout, Json
+STD_OUTPUT = 'stdout'
+WINDOW_OUTPUT = 'window'
+IMAGE_OUTPUT = 'image_output%04d.jpg'
+VIDEO_OUTPUT = 'video_output.mp4'
+JSON_OUTPUT = 'test_output%04d.json'
```
leo-p: What about bulk json?

deepomatic/cli/thread_base.py
```diff
@@ -1,52 +1,254 @@
-import threading
-import time
 import logging
 import traceback
+import gevent
+import signal
+from contextlib import contextmanager
+from gevent.threadpool import ThreadPool
+from threading import Lock
```
leo-p: Why not use [gevent semaphore](https://sdiehl.github.io/gevent-tutorial/#locks-and-semaphores)?

deepomatic/cli/thread_base.py
```diff
@@ -1,52 +1,254 @@
...
 
-POP_TIMEOUT = 1
+try:
+    from Queue import Empty, Full, Queue, LifoQueue
```
leo-p: Same why not use gevent [queues](https://sdiehl.github.io/gevent-tutorial/#queues)?

deepomatic/cli/thread_base.py
```diff
@@ -1,52 +1,254 @@
...
+                        self.stop()
+                        return
+            if not self.alive:
+                return
+
+    def process_msg(self, msg):
         raise NotImplementedError()
 
+    def pop_input(self):
+        return self.input_queue.get(block=False)
```
leo-p: Equivalent to `gevent` `get_nowait` ?

deepomatic/cli/thread_base.py
```diff
@@ -1,52 +1,254 @@
...
+
+    def put_to_output(self, msg_out):
+        while True:
+            try:
+                self.output_queue.put(msg_out, block=False)
```
leo-p: Equivalent to `gevent` `put_nowait` ?

deepomatic/cli/thread_base.py
```diff
@@ -1,52 +1,254 @@
...
+class Pool(object):
+    def __init__(self, nb_thread, thread_cls=Thread, thread_args=(), thread_kwargs=None, name=None):
+        self.nb_thread = nb_thread
+        self.name = name or thread_cls.__name__
+        self.threads = []
+        self.thread_cls = thread_cls
+        self.thread_args = thread_args
+        self.thread_kwargs = thread_kwargs or {}
+
+    def start(self):
```
leo-p: Use [gevent groups](https://sdiehl.github.io/gevent-tutorial/#groups-and-pools) ?

deepomatic/cli/output_data.py
```diff
@@ -2,27 +2,34 @@
...
 def save_json_to_file(json_data, json_path):
     try:
-        with open('%s.json' % json_path, 'w') as file:
-            logging.info('Writing %s.json' % json_path)
-            json.dump(json_data, file)
+        with open('%s.json' % json_path, 'w') as f:
+            LOGGER.info('Writing %s.json ..' % json_path)
+            json.dump(json_data, f)
+            # force writing directly on the disk
+            # or it might be a bit delayed and makes sometimes the files
```
leo-p: Sentence not finished?

deepomatic/cli/output_data.py
```diff
@@ -116,18 +128,18 @@ def __init__(self, descriptor, **kwargs):
         self._i = 0
 
     def output_frame(self, frame):
+        self._i += 1
```
leo-p: This makes the frame start at `1`. Is this what we want?

deepomatic/cli/input_data.py
```diff
@@ -2,21 +2,19 @@
...
 from .workflow import get_workflow
 from .output_data import OutputThread
-try:
-    from Queue import Queue, LifoQueue, Empty
-except ImportError:
-    from queue import Queue, LifoQueue, Empty
 
 
-QUEUE_MAX_SIZE = 50
+# note that it could probably be dynamically calculated
```
leo-p: What is this referencing? Artifact from `QUEUE_MAX_SIZE`?

deepomatic/cli/cmds/feedback.py
```diff
@@ -2,26 +2,32 @@
...
-from .studio_helpers.file import File
+from .studio_helpers.file import DatasetFiles, UploadImageGreenlet
 from .studio_helpers.task import Task
+from ..common import TqdmToLogger
+from ..thread_base import Queue, Empty, Greenlet, Pool, Thread, MainLoop, QUEUE_MAX_SIZE
+
 
 ###############################################################################
 
+GREENLET_NUMBER = 10
```
leo-p: What is the rule of thumb here? Does it have to be roughly equal to the number of CPU cores?

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -3,79 +3,66 @@
...
-try:
-    from Queue import Queue, Empty
-except ImportError:
-    from queue import Queue, Empty
+from ...thread_base import Queue, Empty, Greenlet, Pool
 
 
-THREAD_NUMBER = 5
+LOGGER = logging.getLogger(__name__)
+GREENLET_NUMBER = 10
```
leo-p: This parameter is referenced twice, should we make it common?

deepomatic/cli/cmds/feedback.py
```diff
@@ -65,13 +72,38 @@ def main(args):
...
+
     paths = args.get('path', [])
     json_file = args.get('json_file', False)
     recursive = args.get('recursive', False)
 
     # Scan to find all files
     files = get_all_files(paths=paths, find_json=json_file, recursive=recursive)
 
+    # TODO: add a maxsize to avoid taking too much memories
+    # This implied reading twice to get the total_files
```
leo-p:
```suggestion
    # This implies reading twice to get the total_files
```

tests/utils.py
maingoh: input is a keyword/global python function, that is why I named it this way

tests/utils.py
maingoh: It is currently not tested but yes I can add it in the if

tests/utils.py
maingoh: It is not tested, and wasn't, so we can add this test in another PR

deepomatic/cli/thread_base.py
maingoh: I think the gevent semaphore is just to synchronize greenlets, I wanted a lock that works with thread also. Though I am not totally sure of the difference between both

deepomatic/cli/thread_base.py
maingoh: It didn't work with threads (in my first try). 
I had a hard time finding the synchronization objects and making everything working together with a graceful shutdown, so I think I shouldn't touch more in this PR.
The main thing is that using gevent blocking syncronization objects will yield to other greenlets when waiting, but here we don't need to wait

deepomatic/cli/thread_base.py
maingoh: it exists in Queue also but I prefer using the flag

deepomatic/cli/thread_base.py
maingoh: The purpose of this class is to abstract the executor (thread or greenlets), so I want both behavior to be the same, If I use gevents groups, I would need to make two types of Pool which I wanted to avoid.

deepomatic/cli/output_data.py
maingoh: Probably yes, though I don't think the fsync is still needed. I will remove it to see if tests are passing. I think the bug was something else when I added this

deepomatic/cli/output_data.py
maingoh: The main problem here is that JsonOutputData use filename starting at 1 and ImageOutputData use a filename starting at 0. So it is not synchronized and would lead to bugs. We can start both at 0, but do we want a filename named `image_0000.jpg` ?

deepomatic/cli/cmds/studio_helpers/file.py
maingoh: It is actually only needed here. Both pool don't perform the same operations so for now it is probably better to fix the number of greenlets separately.

deepomatic/cli/cmds/feedback.py
maingoh: There is no real rule. This depends on the scenario (CPU and type of I/O). We could put more, but we don't want to DDOS the API. Also depending on your CPU it might get slower.
In general we could calculate this dynamically but it is hard to do it right.

maingoh:
The main thing to understand with gevent is that all greenlets runs in the main thread. That means if one greenlet blocks without I/O for a "long" time, it will block all other greenlets. So basically when doing Image processing (decoding / encoding / drawing / bluring) it is quite CPU intensive and thus cannot be done in the a greenlet (or it will impact the overall throughput). 
There are many possible optimizations, but without performance benchmarks it is hard to find the best combination of thread / greenlets and have real insights, this will be my main focus (I have a todo of things to tests later).

deepomatic/cli/input_data.py
```diff
@@ -57,78 +59,65 @@ def get_input(descriptor, kwargs):
...
+            frame = next(self.inputs)
+        except StopIteration:
+            self.stop()
+            return
+
+        frame.frame_number = self.frame_number
+        if self.inputs.is_infinite():
+            # Discard all previous inputs
+            with self.output_queue.mutex:
+                self.output_queue.clear()
```
alemoreau: Lifo Queue has no method `clear`

alemoreau:
```suggestion
                while True:
                    try:
                        self.output_queue.get()
                        self.output_queue.task_done()
                    except Empty:
                        break
                    
```

maingoh: This is much slower, I would prefer keeping the first version and eventually doing something else for the LifoQueue, but do we really want to keep the LifoQueue as we discussed before ?

maingoh: I thought you were commenting in clear_queues methods. 
Actually LifoQueue use a list internally which has not method clear in python2 only. 
We can do:
```
def clear_queue(queue):
    with queue.mutex:
        if isinstance(queue, LifoQueue):
             queue.queue = []
       else:
             queue.queue.clear()
```

deepomatic/cli/output_data.py
leo-p: As long as it's coherent between `JSONOutputData` and `ImageOutputData` it should be ok.

PR #52, opened 2019-04-10
Improves drawing

leo-p:
- Improves the drawing function for tag and detection

- Tagging
draw_tag

- Detection
draw_det

### DIFF
--- a/deepomatic/cli/cmds/infer.py
+++ b/deepomatic/cli/cmds/infer.py
@@ -31,8 +31,7 @@ class DrawImagePostprocessing(object):
     def __call__(self, frame):
         frame.output_image = frame.image.copy()
         output_image = frame.output_image
-        height = output_image.shape[0]
-        width = output_image.shape[1]
+        height, width, _ = output_image.shape
         tag_drawn = 0  # Used to store the number of tags already drawn
         for pred in frame.predictions['outputs'][0]['labels']['predicted']:
             # Build legend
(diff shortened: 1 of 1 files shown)


deepomatic/cli/cmds/infer.py
```diff
@@ -21,34 +31,68 @@ def __init__(self, **kwargs):
...
+            if self._draw_scores:
+                label += str(round(pred['score'], 4))
+            # Make sure labels are ascii
+            label = label.encode('ascii', 'replace')
+
+            # Get text draw parameters
+            ret, baseline = cv2.getTextSize(label, cv2.FONT_HERSHEY_SIMPLEX, FONT_SCALE, SCORE_DECIMAL_PRECISION)
+
+            # If we have a bounding box
+            if 'roi' in pred:
```
maingoh: I prefer with 
```
 roi = pred.get('roi')
 if roi is not None:
```
This avoid duplicating the rewriting twice the key and handle `if pred['roi'] is None`

deepomatic/cli/cmds/infer.py
```diff
@@ -21,34 +31,68 @@ def __init__(self, **kwargs):
...
                 # Retrieve coordinates
-                bbox = roi['bbox']
-                xmin = int(bbox['xmin'] * w)
-                ymin = int(bbox['ymin'] * h)
-                xmax = int(bbox['xmax'] * w)
-                ymax = int(bbox['ymax'] * h)
+                bbox = pred['roi']['bbox']
+                xmin = int(bbox['xmin'] * width)
```
maingoh: can you make a function for that returns the absolute cordinates of the pred and use it in blur as well

deepomatic/cli/cmds/infer.py
```diff
@@ -21,34 +31,68 @@ def __init__(self, **kwargs):
     def __call__(self, frame):
         frame.output_image = frame.image.copy()
         output_image = frame.output_image
-        h = output_image.shape[0]
-        w = output_image.shape[1]
+        height, width, _ = output_image.shape
```
alemoreau: We should be careful later, if we use cv2.imread(file_path, 0) (grayscale), this will crash

PR #53, opened 2019-04-11
Fix directory input and window output

alemoreau:
- There are some outputs expecting sorted inputs (video, window...), so for unordered input (directory), the output looks weird
- window output was not working

### DIFF
--- a/deepomatic/cli/input_data.py
+++ b/deepomatic/cli/input_data.py
@@ -271,7 +271,7 @@ class DirectoryInputData(InputData):
         if self.is_valid(descriptor):
             _paths = [os.path.join(descriptor, name) for name in os.listdir(descriptor)]
             self._inputs = []
-            for path in _paths:
+            for path in sorted(_paths):
                 if ImageInputData.is_valid(path):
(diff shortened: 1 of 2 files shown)


deepomatic/cli/input_data.py
```diff
@@ -271,7 +271,7 @@ def __init__(self, descriptor, **kwargs):
         if self.is_valid(descriptor):
             _paths = [os.path.join(descriptor, name) for name in os.listdir(descriptor)]
             self._inputs = []
-            for path in _paths:
+            for path in sorted(_paths):
```
leo-p: I'm not sure I understand why we need this.

alemoreau: When using a video output for instance, you expect the frames to be ordered. But here _paths contains the image paths in an unclear order (maybe by modification date ? I don't know) so the video looks weird (it jumps randomly from frame to frame). Usually the names of the files in a directory represent their order, (frame_00.jpg, frame_01.jpg.. etc), so by sorting by name, we can produce a video in the same order

PR #57, opened 2019-04-15
Proper label string encoding

leo-p:


alemoreau:
Care, unidecode behaves differently with python2

> The module exports a function that takes an Unicode object (Python 2.x) or string (Python 3.x)
https://pypi.org/project/Unidecode/

See tests for python2.7
`=============================== warnings summary ===============================
tests/test_draw.py::test_e2e_image_draw_image
  /home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/unidecode/__init__.py:46: RuntimeWarning: Argument <type 'str'> is not an unicode object. Passing an encoded string will likely have unexpected results.
    _warn_if_not_unicode(string)`

leo-p:
Removed unidecode for license reasons.

To do: Handle special characters Issue https://github.com/Deepomatic/deepocli/issues/58

deepomatic/cli/cmds/infer.py
```diff
@@ -45,15 +46,16 @@ def __call__(self, frame):
...
             if self._draw_labels:
                 label = pred['label_name']
             if self._draw_labels and self._draw_scores:
                 label += ' '
             if self._draw_scores:
                 label += str(round(pred['score'], SCORE_DECIMAL_PRECISION))
+            label = unidecode(label)
             # Make sure labels are ascii
-            label = str(label.encode('ascii', 'replace'))
+            label = label.encode('ascii', 'replace').decode('ascii')
```
maingoh: you shouldn't need this line, the purpose of unidecode is to return ascii

maingoh: if you can also put a comment to explain why we need to unidecode (it is not supported by either opencv or the font)

PR #59, opened 2019-04-15
Harmonizes fps for i/o videos

leo-p:
The retained behavior is the following:
- If an `fps` argument is provided:
    - Input: the frame are extracted according to the `fps` value.
    - Output: the video is reconstructed according to the `fps` value.
- If not `fps` argument is provided:
    - If the input is a video, its `fps` value is used.
    - If the input is not a video, the `DEFAULT_FPS` value is used (for instance when outputing a video from images).

### DIFF
--- a/deepomatic/cli/input_data.py
+++ b/deepomatic/cli/input_data.py
@@ -60,10 +60,6 @@ def get_input(descriptor, kwargs):
 def input_loop(kwargs, postprocessing=None):
     inputs = get_input(kwargs.get('input', 0), kwargs)
 
-    # If no fps is specified and input is a video, pass the fps parameter to the output
-    if isinstance(inputs, VideoInputData):
-        kwargs['fps'] = inputs.get_fps()
-
     # Initialize progress bar
     logging.basicConfig(level=logging.INFO)
     logger = logging.getLogger()
@@ -214,6 +210,7 @@ class VideoInputData(InputData):
         self._i = 0
         self._name = '%s_%s_%s' % (self._name, '%05d', self._reco)
         self._cap = cv2.VideoCapture(self._descriptor)
+        self._kwargs_fps = kwargs['input_fps']
         self._fps = self.get_fps()
 
     def __iter__(self):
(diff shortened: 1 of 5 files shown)


deepomatic/cli/input_data.py
```diff
@@ -60,6 +60,10 @@ def get_input(descriptor, kwargs):
 def input_loop(kwargs, postprocessing=None):
     inputs = get_input(kwargs.get('input', 0), kwargs)
 
+    # If no fps is specified and input is a video, pass the fps parameter to the output
+    if isinstance(inputs, VideoInputData):
+        kwargs['fps'] = inputs.get_fps()
```
alemoreau: What about the `If no fps is specified` condition ?

leo-p: Then `DEFAULT_FPS` kicks in here https://github.com/Deepomatic/deepocli/blob/1362263978f7d2967ff0859364b251ae532adfe1/deepomatic/cli/output_data.py#L16
https://github.com/Deepomatic/deepocli/blob/1362263978f7d2967ff0859364b251ae532adfe1/deepomatic/cli/output_data.py#L149

alemoreau: What I meant is that line 64 only checks that the input is a video, and if it is, it will overwrite the given fps with the input fps (even if the user specified an output fps).
So you won't be able to set the output fps in case of video input, right ?

alemoreau: So you probably want to do like the comment l.63 describes, checking that no fps is specified and that the input is a video before overwriting the user-specified fps

maingoh:
I think we should also makes sure `self._cap.get(cv2.CAP_PROP_FPS)` doesn't fail (returns something higher than 0 ?). And also how do we handle streams/webcam ? I think we can also set a FPS at the VideoCapture level. Probably better than doing the adjusting ourself ?

deepomatic/cli/input_data.py
alemoreau: weird stuffs happening here : https://github.com/Deepomatic/deepocli/blob/1362263978f7d2967ff0859364b251ae532adfe1/deepomatic/cli/input_data.py#L214

leo-p: Your comment made me realize that we might want to always overwrite the cli `fps` with the input video value if it's used.

The `fps` value passed as a parameter is not the value used in the video input, see here:
https://github.com/Deepomatic/deepocli/blob/1362263978f7d2967ff0859364b251ae532adfe1/deepomatic/cli/input_data.py#L215

We have to choose between two behaviors:
- 1) Enforce the actual same fps value for input and output
- 2) Use the same fps value for input and output initialization, which might lead to cases where the fps is capped for input and not for output thus leading to sped-up output videos.

I have a preference for 1). What do you think?

alemoreau:
This looks complicated.
self._cap.get(cv2.CAP_PROP_FPS) usually fails for webcam and streams
https://github.com/Deepomatic/deepocli/blob/1362263978f7d2967ff0859364b251ae532adfe1/deepomatic/cli/input_data.py#L214

deepomatic/cli/input_data.py
```diff
@@ -246,6 +263,40 @@ def __next__(self):
...
+        #   _video_fps: original video fps
+        #   _kwarg_fps: fps specified by the user through the CLI if any
+        #   _extract_fps: fps used for frame extraction
+
+        if self._cap is not None:
+            # Retrieve the original video fps if available
+            try:
+                self._video_fps = self._cap.get(cv2.CAP_PROP_FPS)
+            except Exception:
+                raise ValueError('Could not read fps for video {}, please specify it with --fps option.'.format(self._descriptor))
```
maingoh: typo `--input_fps`

deepomatic/cli/input_data.py
```diff
@@ -246,6 +263,40 @@ def __next__(self):
...
+            if self._video_fps == 0:
+                raise ValueError('Null fps detected for video {}, please specify it with --fps option.'.format(self._descriptor))
+            
+            # Compute fps for frame extraction so that we don't analyze useless frame that will be discarded later
+            if not self._kwargs_fps:
+                self._extract_fps = self._video_fps
+                logging.info('No keyword fps specified, using raw video fps of {}'.format(self._video_fps))
```
maingoh: No `--input_fps` specified

deepomatic/cli/input_data.py
```diff
@@ -246,6 +263,40 @@ def __next__(self):
...
+            elif self._kwargs_fps < self._video_fps:
+                self._extract_fps = self._kwargs_fps
+                logging.info('Using keyword fps of {} instead of raw video fps of {}'.format(self._kwargs_fps, self._video_fps))
```
maingoh: Using user specified `--input_fps`

deepomatic/cli/output_data.py
```diff
@@ -146,7 +146,7 @@ def __init__(self, descriptor, **kwargs):
         elif ext == '.mp4':
             fourcc = cv2.VideoWriter_fourcc('m', 'p', '4', 'v')
         self._fourcc = fourcc
-        self._fps = kwargs['fps'] if kwargs['fps'] else DEFAULT_FPS
+        self._fps = kwargs['output_fps'] if kwargs['output_fps'] else DEFAULT_FPS
```
maingoh: You put the whole logic in input_loop, so let's add this there also (with a log when using DEFAULT_FPS)

deepomatic/cli/output_data.py
```diff
@@ -181,7 +181,7 @@ def output_frame(self, frame):
 class DisplayOutputData(OutputData):
     def __init__(self, **kwargs):
         super(DisplayOutputData, self).__init__(None, **kwargs)
-        self._fps = kwargs.get('fps', DEFAULT_FPS)
+        self._fps = kwargs['output_fps'] if kwargs['output_fps'] else DEFAULT_FPS
```
maingoh: same here

PR #72, opened 2019-06-19
Reuse prediction for draw and infer

leo-p:
Adds the `--from_file` option to reuse previously computed predictions with `infer` and `blur`
```bash
deepo infer -i video.mp4 -o predictions.json -r 123
deepo blur -i video.mp4 -o video_blurred.mp4 -r 123 --from_file predictions.json
```

Also simplifies errors display and implements JSON schema for validation

### DIFF
--- a/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
+++ b/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
@@ -1,44 +1,47 @@
 
-def transform_json_from_vulcan_to_studio(vulcan_json, name, filename):
+def transform_json_from_vulcan_to_studio(vulcan_json):
     """Transforms a json from the vulcan format to the Studio format."""
     # Initialize variables
     studio_json = {'tags': [], 'images': []}
     unique_tags = set()
 
-    # Loop through all images
-    img_studio = {'annotated_regions': []}
-    all_predictions = vulcan_json['outputs'][0]['labels']['predicted']
-    for prediction in all_predictions:
-        # Build studio annotation
-        annotation = {
-            "tags": [prediction['label_name']],
-            "region_type": "Whole",
-            "score": prediction['score'],
-            "threshold": prediction['threshold']
-        }
+    # Transform vulcan json to a list of vulcan json if needed
+    if isinstance(vulcan_json, dict):
+        vulcan_json = [vulcan_json]
 
-        # Add bounding box
-        if 'roi' in prediction:
-            annotation['region_type'] = 'Box'
-            annotation['region'] = {
-                "xmin": prediction['roi']['bbox']['xmin'],
-                "xmax": prediction['roi']['bbox']['xmax'],
-                "ymin": prediction['roi']['bbox']['ymin'],
-                "ymax": prediction['roi']['bbox']['ymax']
+    # Loop through all vulcan images
+    for vulcan_image in vulcan_json:
+        # Initialize studio images
+        studio_image = {'annotated_regions': []}
+
+        # Loop through all vulcan predictions
+        for prediction in vulcan_image['outputs'][0]['labels']['predicted']:
+            # Build studio annotation in case of classification or tagging
+            annotation = {
+                "tags": [prediction['label_name']],
+                "region_type": "Whole",
+                "score": prediction['score'],
+                "threshold": prediction['threshold']
             }
 
-        # Update json
-        img_studio['annotated_regions'].append(annotation)
+            # Add bounding box if needed
+            if 'roi' in prediction:
+                annotation['region_type'] = 'Box'
+                annotation['region'] = {
+                    "xmin": prediction['roi']['bbox']['xmin'],
+                    "xmax": prediction['roi']['bbox']['xmax'],
+                    "ymin": prediction['roi']['bbox']['ymin'],
+                    "ymax": prediction['roi']['bbox']['ymax']
+                }
 
-        # Update unique tags
-        unique_tags.add(prediction['label_name'])
+            # Update json and unique tags
+            studio_image['annotated_regions'].append(annotation)
+            unique_tags.add(prediction['label_name'])
 
-    # Update json vesta
-    studio_json['images'].append(img_studio)
-    studio_json['images'][0]['location'] = name
-    studio_json['images'][0]['data'] = {'filename': filename}
+        # Update studio json
+        studio_json['images'].append(studio_image)
 
-    # Update unique tags
+    # Update final unique tags
     studio_json['tags'] = list(unique_tags)
 
     return studio_json
--- a/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
+++ b/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
@@ -12,7 +12,9 @@ def transform_json_from_vulcan_to_studio(vulcan_json):
     # Loop through all vulcan images
     for vulcan_image in vulcan_json:
         # Initialize studio images
-        studio_image = {'annotated_regions': []}
+        studio_image = {'annotated_regions': [], 'location': vulcan_image.get('location', '')}
+        if 'data' in vulcan_image:
+            studio_image['data'] = vulcan_image['data']
 
         # Loop through all vulcan predictions
         for prediction in vulcan_image['outputs'][0]['labels']['predicted']:
--- a/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
+++ b/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
@@ -59,6 +59,8 @@ def transform_json_from_studio_to_vulcan(studio_json):
     for studio_image in studio_json['images']:
         # Initialize vulcan prediction
         vulcan_pred = {'outputs': [{'labels': {'discarded': [], 'predicted': []}}]}
+        predicted = []
+        discarded = []
         for metadata in ['location', 'data']:
             if metadata in studio_image:
                 vulcan_pred[metadata] = studio_image[metadata]
@@ -85,9 +87,15 @@ def transform_json_from_studio_to_vulcan(studio_json):
 
             # Update json
             if annotation['score'] >= annotation['threshold']:
-                vulcan_pred['outputs'][0]['labels']['predicted'].append(annotation)
+                predicted.append(annotation)
             else:
-                vulcan_pred['outputs'][0]['labels']['discarded'].append(annotation)
+                discarded.append(annotation)
+
+        # Sort by prediction score of descending order
+        predicted = sorted(predicted, key=lambda k: k['score'], reverse=True)
+        discarded = sorted(discarded, key=lambda k: k['score'], reverse=True)
+        vulcan_pred['outputs'][0]['labels']['predicted'] = predicted
+        vulcan_pred['outputs'][0]['labels']['discarded'] = discarded
 
         # Update vulcan json
         vulcan_json.append(vulcan_pred)
--- a/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
+++ b/deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
@@ -1,6 +1,6 @@
 
 def transform_json_from_vulcan_to_studio(vulcan_json):
-    """Transforms a json from the vulcan format to the Studio format."""
+    """Transforms a json from the vulcan format to the studio format."""
     # Initialize variables
     studio_json = {'tags': [], 'images': []}
     unique_tags = set()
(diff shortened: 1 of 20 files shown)


deepomatic/cli/workflow/json_workflow.py
```diff
@@ -0,0 +1,55 @@
...
+    def close(self):
+        pass
+
+    def infer(self, _useless_encoded_image_bytes, _useless_push_client, frame_name):
+        # _useless_encoded_image_bytes and _useless_push_client are used only for rpc and cloud workflows
+        try:
+            frame_pred = self._all_predictions[frame_name]
+        except KeyError:
+            raise InferenceError("Could not find predictions for frame {}".format(frame_name))
+        except:
```
maingoh: useless

deepomatic/cli/workflow/json_workflow.py
```diff
@@ -0,0 +1,55 @@
...
+
+    class InferResult(AbstractWorkflow.AbstractInferResult):
+        def __init__(self, frame_name, frame_pred):
+            self.frame_name = frame_name
+            self.frame_pred = frame_pred
+
+        def get_predictions(self, timeout):
+            return self.frame_pred
+
+    def __init__(self, recognition_version_id, pred_file):
```
maingoh: I think the `recognition_version_id` is not needed with a `--from-file`

deepomatic/cli/output_data.py
```diff
@@ -266,12 +267,10 @@ def close(self):
     def output_frame(self, frame):
         self._i += 1
         predictions = frame.predictions
+        predictions['location'] = frame.name
+        predictions['data'] = {'filename': frame.filename}
```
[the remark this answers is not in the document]
maingoh: can you wrap the predictions inside a `predictions` key ? I would like the json inside this key to be exactly a vulcan json.

deepomatic/cli/workflow/json_workflow.py
```diff
@@ -0,0 +1,55 @@
...
+            vulcan_json_with_pred = json.load(json_file)
+        # Transform it to vulcan format if needed
+        if validate_vulcan_json(pred_file):
+            LOGGER.debug("Vulcan prediction JSON {} validated".format(pred_file))
+        if validate_studio_json(pred_file):
+            vulcan_json_with_pred = transform_json_from_studio_to_vulcan(vulcan_json_with_pred)
+            LOGGER.debug("Studio prediction JSON {} validated and transformer to Vulcan format".format(pred_file))
+
+        # Store predictions for easy access
+        self._all_predictions = {vulcan_pred['location']: vulcan_pred for vulcan_pred in vulcan_json_with_pred}
```
maingoh: Is it working in python2 ?

deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
```diff
@@ -1,44 +1,95 @@
...
+                        'xmin': studio_pred['region']['xmin'],
+                        'xmax': studio_pred['region']['xmax'],
+                        'ymin': studio_pred['region']['ymin'],
+                        'ymax': studio_pred['region']['ymax']
+                    }
+                }
+
+            # Update json
+            if annotation['score'] >= annotation['threshold']:
+                vulcan_pred['outputs'][0]['labels']['predicted'].append(annotation)
```
maingoh: They should be sorted. Is it always the case ?

deepomatic/cli/workflow/json_workflow.py
```diff
@@ -0,0 +1,55 @@
...
+        # Load json
+        with open(pred_file) as json_file:
+            vulcan_json_with_pred = json.load(json_file)
+        # Transform it to vulcan format if needed
+        if validate_vulcan_json(pred_file):
+            LOGGER.debug("Vulcan prediction JSON {} validated".format(pred_file))
+        if validate_studio_json(pred_file):
+            vulcan_json_with_pred = transform_json_from_studio_to_vulcan(vulcan_json_with_pred)
+            LOGGER.debug("Studio prediction JSON {} validated and transformer to Vulcan format".format(pred_file))
+
```
maingoh: If both fail, raise ?

deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
leo-p: Sorted by prediction score?

deepomatic/cli/input_data.py
```diff
@@ -13,34 +14,51 @@
...
+from .cmds.studio_helpers.vulcan2studio import transform_json_from_vulcan_to_studio
 
 
 LOGGER = logging.getLogger(__name__)
 
 
 def get_input(descriptor, kwargs):
     if descriptor is None:
-        raise NameError('No input specified. use -i flag')
+        LOGGER.error('No input specified. use -i flag')
```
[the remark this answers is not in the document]
leo-p: Duly noted, will roll back and catch exception in main routine.

deepomatic/cli/workflow/json_workflow.py
leo-p: Including `recognition_version_id` has two advantages: 1) Making sure the predictions used from the json are actually what we are expecting 2) Harmonizing the behaviour with RPC where we also specify the `recognition_version_id`

deepomatic/cli/input_data.py
```diff
@@ -341,10 +365,16 @@ def __init__(self, descriptor, **kwargs):
             self._inputs = []
             for path in sorted(_paths):
                 if ImageInputData.is_valid(path):
+                    LOGGER.debug('Image input data detected for {}'.format(path))
                     self._inputs.append(ImageInputData(path, **kwargs))
                 elif VideoInputData.is_valid(path):
+                    LOGGER.debug('Video input data detected for {}'.format(path))
                     self._inputs.append(VideoInputData(path, **kwargs))
+                elif JsonInputData.is_valid(path):
```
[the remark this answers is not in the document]
leo-p: This is just a studio JSON with image location.

deepomatic/cli/input_data.py
```diff
@@ -408,63 +438,43 @@ class JsonInputData(InputData):
...
-                    None for path in _paths if os.path.isfile(path)]
-                self._inputs = [_input for _input in _files if _input is not None]
+                studio_json = json.load(json_file)
+
+        # Go through all locations and check input validity
+        self._inputs = []
+        for studio_img in studio_json['images']:
+            img_location = studio_img['location']
+            # If the file does not exist, ignore it
+            if not os.path.isfile(img_location):
```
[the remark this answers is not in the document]
leo-p: For the moment no, we only handle local files, this could be an improvement in the future.

deepomatic/cli/workflow/json_workflow.py
leo-p: Yes!

deepomatic/cli/json_schema.py
```diff
@@ -0,0 +1,144 @@
+import os
+import sys
+import json
+import logging
+from jsonschema import validate
```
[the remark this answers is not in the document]
leo-p: Good to know for next time or if the json schema changes!

deepomatic/cli/output_data.py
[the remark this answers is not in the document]
leo-p: `data` makes it coherent between `studio` and `vulcan` json format.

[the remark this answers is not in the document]
leo-p: For the `predictions` key I think that would defeat the whole purpose, this is the final JSON that will be stored on disk. We want it to be as close as possible to a Vulcan JSON, no?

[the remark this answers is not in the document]
maingoh: ok for `data` even though it is a bit too generic. For the rest my point is that I wanted to distinct the initial run format from the one generated by deepocli. You added extra keys, so if you want to compare this json to another file that contains run predictions not saved by deepocli (without the extra keys), it will be less convenient. Though I guess you can keep it like that for now.

deepomatic/cli/cmds/studio_helpers/vulcan2studio.py
maingoh: Yes both `predicted` and `discarded` should be sorted by prediction score. This is probably not the case when converting from a studio json ?

deepomatic/cli/workflow/json_workflow.py
maingoh: How do you make sure ? It is not saved in the prediction file. I could always put `--recognition_version_id 123` and it would work even though it doesn't exist or I don't have access. This is just more confusing, because you think it is needed but it is not. If you really want to add this option, please add it to the prediction file, and makes sure they match.

deepomatic/cli/input_data.py
```diff
@@ -128,7 +145,14 @@ def input_loop(kwargs, postprocessing=None):
     ]
 
     loop = MainLoop(pools, queues, pbar, lambda: workflow.close())
-    stop_asked = loop.run_forever()
+    try:
+        stop_asked = loop.run_forever()
+    except BadStatus as e:
```
maingoh: I think it is never triggered here, it is only triggered in threads. I will fix this.

deepomatic/cli/json_schema.py
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+}
+
+
+def validate_json(json_path, json_schema):
+    """Validate a JSON using a schema"""
+    # Check if the file exists
+    if not os.path.isfile(json_path):
+        LOGGER.error('File {} does not exists')
```
maingoh: Probably raising an exception is better, (you can log the error later). Just let this function crash, and where you try both format you can store them temporarily and log them only if both checks fail.

deepomatic/cli/workflow/json_workflow.py
```diff
@@ -0,0 +1,54 @@
...
+        def get_predictions(self, timeout):
+            return self.frame_pred
+
+    def __init__(self, recognition_version_id, pred_file):
+        super(JsonRecognition, self).__init__('r{}'.format(recognition_version_id))
+        self._id = recognition_version_id
+        self._pred_file = pred_file
+
+        # Check json validity
+        if not(validate_studio_json(pred_file)) and not(validate_vulcan_json(pred_file)):
```
maingoh: You validate multiple times, why not simply:
```
 if validate_vulcan_json(pred_file):
     ...
 elif validate_studio_json(pred_file):
    ...
 else:
        raise DeepoPredictionJsonError("Prediction JSON file {} is neither a proper Studio or Vulcan JSON file".format(pred_file))


```

deepomatic/cli/workflow/json_workflow.py
leo-p: We identify predictions using `frame.name` which is composed of:
- the `filename`
- a sequence number for videos
- the `recognition_id`.
That's also what is stored in any prediction json.

Thus if you were to use `--recognition_version_id 123` with file `img.png` we would look for the prediction associated with `frame.name` `img_123` in the json and not find it.

deepomatic/cli/input_data.py
maingoh: Done here https://github.com/Deepomatic/deepocli/pull/72/commits/0e438aea7ef34d049e7ecc20cb6ad774cbd2c1a7

deepomatic/cli/input_data.py
```diff
@@ -291,20 +312,20 @@ def get_fps(self):
...
-            raise ValueError('Could not read fps for video {}, please specify it with --input_fps option.'.format(self._descriptor))
+            raise DeepoFPSError('Could not read fps for video {}, please specify it with --input_fps option.'.format(self._descriptor))
         if self._video_fps == 0:
-            raise ValueError('Null fps detected for video {}, please specify it with --input_fps option.'.format(self._descriptor))
+            raise DeepoFPSError('Null fps detected for video {}, please specify it with --input_fps option.'.format(self._descriptor))
 
         # Compute fps for frame extraction so that we don't analyze useless frame that will be discarded later
         if not self._kwargs_fps:
             self._extract_fps = self._video_fps
-            LOGGER.info('No --input_fps specified, using raw video fps of {}'.format(self._video_fps))
```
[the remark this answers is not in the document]
leo-p: This is the internal function debug log, but there is also an [informative prompt](https://github.com/Deepomatic/deepocli/pull/72/files/a443b55f8777a6f6d211e52ca6c5a04e1c705855#diff-d1ea71c2f8fe6444c402cf06265aa68fR110) displayed to the user that takes into consideration the `output_fps` too and all the possible combinations.

Using the `info` level here sometimes leads to double or misleading messaging.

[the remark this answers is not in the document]
maingoh: Agreed those messages should be displayed to the user

PR #70, opened 2019-06-17
Import Images Preprocessing

romyli:
Do not merge this PR before studio release v1.20.0 (end of month)

Change deepocli feedback code to allow importing images with json containing video preprocessing parameters : 
```
{
    "videos": [
        {
            "location": "path",
            "preprocessing":{"fps":"1"},
            "data": {
                "video": "BAY 3_20181231112829",
                "object": "None"
            }
        },
        {
            "location": "video_file",
            "preprocessing":{"fps":"3", "start_time": "0:01:00.0"}
        }
    ],
    "images": [
        {
            "location": "image_path"
        }
    ]
}
```

PR #78, opened 2019-07-11
Support for vulcan json upload

leo-p:
- Support Vulcan format JSON for image upload to Studio.
- Detect automatically the JSON type.
- Implement JSON schema validation.
- Replace `location` in infer JSON with real location and add metadata in `data` field.


deepomatic/cli/cmds/infer.py
```diff
@@ -155,8 +153,11 @@ def close(self):
         self.workflow.close_client(self.push_client)
 
     def process_msg(self, frame):
-        frame.inference_async_result = self.workflow.infer(frame.buf_bytes, self.push_client, frame.name)
-        return frame
+        try:
+            frame.inference_async_result = self.workflow.infer(frame.buf_bytes, self.push_client, frame.name)
+            return frame
+        except InferenceError as e:
+            LOGGER.error('Error getting predictions for frame {}'.format(frame))
```
maingoh: - display `e`, it contains some important information
- can you return explicitely `None`

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -72,60 +76,65 @@ def fill_queue(self, files, dataset_name, commit_pk):
...
-
-                # If it's a type-1 JSON, transform it into a type-2 JSON
-                if 'location' in json_objects:
-                    obj_extension = json_objects['location'].split('.')[-1]
-                    if obj_extension in SUPPORTED_IMAGE_INPUT_FORMAT:
-                        json_objects = {'images': [json_objects]}
-                    elif obj_extension in SUPPORTED_VIDEO_INPUT_FORMAT:
-                        json_objects = {'videos': [json_objects]}
-
+                    raise DeepoOpenJsonError("Upload JSON file {} is not a valid JSON file".format(upload_file))
```
maingoh: This is not true if the file does not exist. Either catch this in the with statement or also catch `IOError`.

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -72,60 +76,65 @@ def fill_queue(self, files, dataset_name, commit_pk):
...
+                # If it's a Studio json, use it directly
+                if is_valid_studio_json(json_data):
+                    LOGGER.warning("Studio JSON {} validated".format(upload_file))
+                    studio_json = json_data
+                # If it's a Vulcan json, transform it to Studio
+                elif is_valid_vulcan_json(json_data):
+                    LOGGER.warning("Vulcan JSON {} validated and transformed to Studio format".format(upload_file))
+                    studio_json = transform_json_from_vulcan_to_studio(json_data)
+                # If it's a Single studio object json, transform it to Studio
+                elif is_valid_single_object_studio_json(json_data):
```
maingoh: I am not sure this is a good idea, why is there another type of format to support ? Two is already enough.

deepomatic/cli/input_data.py
```diff
@@ -116,7 +116,6 @@ def input_loop(kwargs, postprocessing=None):
     pbar = tqdm(total=max_value, file=tqdmout, desc='Input processing', smoothing=0)
 
     # For realtime, queue should be LIFO
-    # TODO: might need to rethink the whole pipeling for infinite streams
```
maingoh: It is not an outdated comment I think, please put it back, it doesn't concern this PR.

deepomatic/cli/output_data.py
```diff
@@ -266,8 +266,11 @@ def close(self):
     def output_frame(self, frame):
         self._i += 1
         predictions = frame.predictions
-        predictions['location'] = frame.name
-        predictions['data'] = {'filename': frame.filename}
+        predictions['location'] = frame.filename
```
maingoh: It is normal you have twice the filename ? You are breaking the location here (it will not match earlier generated location)

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -72,60 +76,65 @@ def fill_queue(self, files, dataset_name, commit_pk):
...
-                    obj_extension = json_objects['location'].split('.')[-1]
-                    if obj_extension in SUPPORTED_IMAGE_INPUT_FORMAT:
-                        json_objects = {'images': [json_objects]}
-                    elif obj_extension in SUPPORTED_VIDEO_INPUT_FORMAT:
-                        json_objects = {'videos': [json_objects]}
-
+                    raise DeepoOpenJsonError("Upload JSON file {} is not a valid JSON file".format(upload_file))
+
+                # If it's a Studio json, use it directly
+                if is_valid_studio_json(json_data):
```
maingoh: You don't already have a function that detect the format and does the conversion somewhere ? We could probably factorize

tests/utils.py
```diff
@@ -1,4 +1,6 @@
 # coding: utf-8
+from gevent.monkey import patch_all
```
maingoh: It is normally already patched earlier, you shouldn't patch here

tests/utils.py
```diff
@@ -8,14 +10,16 @@
...
+# Define outputs
+OUTPUTS = {
+    'STD': 'stdout',
+    'WINDOW': 'window',
+    'IMAGE': 'image_output%04d.jpg',
+    'VIDEO': 'video_output.mp4',
+    'JSON': 'test_output%04d.json',
+    'DIR': 'output_dir'
+}
+OUTPUTS['ALL'] = [output for output in list(OUTPUTS.values()) if output != 'window']
```
maingoh: nit: use global variable instead of hardcoded values `window`

tests/utils.py
```diff
@@ -87,6 +91,32 @@ def check_directory(directory,
                     )
 
 
+def update_path_studio_json(json_pth, image_path):
```
maingoh: Those function all look the same, can't you just have a `load_json_from_file` function and `save_json_to_file` and put them in the same function.
```
def patch_json_for_tests():
    json_data = load_json_from_file()
    json_data['images'][0]['location'] = image_path
    save_json_to_file()

    json_data = load_json_from_file()
     json_data['location'] = image_path
    save_json_to_file()

    json_data = load_json_from_file()
    json_data[0]['location'] = image_path
    save_json_to_file()
```

deepomatic/cli/json_schema.py
```diff
@@ -95,6 +95,45 @@
     }
 }
 
+# Define the single object studio json format
+SINGLE_OBJECT_STUDIO_JSON_SCHEMA = {
```
maingoh: You can't factorize with STUDIO_JSON_SCHEMA ?

deepomatic/cli/cmds/studio_helpers/file.py
leo-p: It depends whether we want to transform to Studio or Vulcan afterwards, plus usually, there are additional cases to deal with. Right now I'm not sure it would simplify the code.

deepomatic/cli/cmds/studio_helpers/file.py
leo-p: I think this format was introduced with the image preprocessing PR #70, if we need to change it this might be better done in another PR.

deepomatic/cli/input_data.py
leo-p: If I remember correctly, I'm the one who wrote this comment, and I don't think it applies anymore because of the Queue refactoring that was done.
Do you think we still need to do additional handling or infinite streams?

deepomatic/cli/output_data.py
leo-p: We are indeed breaking the location but this is voluntary. The goal is to be able to use a json generated with deepocli for a studio upload with deepocli, which was not the case earlier because the `location` key today is not a real location, it's the frame name.

With this PR, the `location` key will always refer to the actual file location and the metadata is used to store both file name and frame name. This way you can chain `infer` and `upload` without having the permute JSON attributes.

tests/utils.py
leo-p: It is not patched when you call the script directly for its `__main__` routine (download all files locally). If we don't patch it, it leads to infinite recursive loops and crashes python.

Either we patch or we move/remove the `__main__` routine. What do you prefer?

deepomatic/cli/input_data.py
maingoh: No it was me. The question was about weather we should use a queue for infinite stream. We probably should not. What we want is juste a single message, that can be overriden by a new one. We don't want a container that accumulate messages, and that we need to flush sometimes.

tests/utils.py
maingoh: Oh ok, I think it doesn't make sense to call this file directly. I would prefer that we remove the `__main__`.

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -96,6 +96,8 @@ def fill_queue(self, files, dataset_name, commit_pk):
                 try:
                     with open(upload_file, 'r') as fd:
                         json_data = json.load(fd)
+                except IOError:
+                    raise DeepoOpenJsonError("Upload JSON file {} does not exist".format(upload_file))
```
maingoh: Ahah `IOError` is not necessarily a not found, it can be a permission denied or something else. I recommend printing the exception instead:
```
except IOError as e:
    raise DeepoOpenJsonError("Upload JSON file failed: {}".format(upload_file, e))
```

tests/test_blur.py
```diff
@@ -101,9 +101,9 @@ def test_e2e_video_blur_video_fps():
     run_blur(INPUTS['VIDEO'], [OUTPUTS['VIDEO']], expect_nb_video=1, extra_opts=['--input_fps', '2'])
 
 
+@pytest.mark.skip(reason="window not handled by test")
```
maingoh: :+1:

PR #82, opened 2019-07-29
Upload object original file name as metadata

leo-p:


romyli:
C'est volontaire de mettre le chemin entier plutôt que juste le nom du fichier?

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -28,9 +28,16 @@ def process_msg(self, msg):
...
             try:
+                # Update file
                 files.update({file['key']: open(file['path'], 'rb')})
-                if 'meta' in file:
-                    meta.update({file['key']: file['meta']})
+
+                # Update corresponding metadata
+                file_meta = file.get('meta', {})
+                if 'data' in file_meta:
+                    file_meta['data']['Original file name'] = file['path']
```
maingoh: - Why do you need to add the filepath in the metadata ? I don't think we want to store the local path of someone's disk
- A dict key shouldn't contain sentences (except maybe for hacky things) or some front end data. Please use lower_case.

Alexis-Jacob: it's when the user want to find an image, it's a recurrent request. But I agree with @romyli we should only keep the filename

leo-p: 1. There is actually a lot of information in the pathname that is lost otherwise. For instance, if you have a `concept1/img1.png` dir schema, once you upload the image we completely lose the `concept1` information. Which usually comes in handy later, but it's too late.
2.  The `data` field is indeed displayed in the front end, that's why the dict key is formatted for human readability. Do you still want to change it to `original_file_name`?

leo-p:
> C'est volontaire de mettre le chemin entier plutôt que juste le nom du fichier?

Yes that's actually the whole point of the PR, see answer to @maingoh .

deepomatic/cli/cmds/studio_helpers/file.py
romyli: Is this feature going to be used essentially for local files (with the recursive option)? We could distinguish this case to add the path starting from the directory used in the command

maingoh: 1. Is it used only for debugging ? Where does this json goes ? It get stored in studio right ? What if a customer upload some files to studio, he probably doesn't want to send us the structure of the disk. If it is a realtive path we could store it, but if it is absolute we might want to try to transform it into a relative path ?
2. What front end ? I meant a front end endpoint that would just take care of sending display data to a templating motor. Here this metadata will be stored into a file right? Please keep lower case and reformat only where needed.

deepomatic/cli/cli_parser.py
```diff
@@ -73,6 +73,7 @@ def argparser_init():
     add_images_parser.add_argument('-o', '--organization', required=True, help="Deepomatic Studio organization slug.", type=str)
     add_images_parser.add_argument('-i', '--input', type=str, nargs='+', required=True, help="One or several input path, either an image or video file (*{}), a directory, or a Studio or Vulcan json (*.json).".format(', *'.join(SUPPORTED_FILE_INPUT_FORMAT)))
     add_images_parser.add_argument('--json', dest='json_file', action='store_true', help='Look for JSON files instead of images.')
+    add_images_parser.add_argument('--show_path', dest='show_path', action='store_true', help='Add the relative path as metadata.')
```
maingoh: `--set_metadata_path` ?

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -15,22 +15,31 @@
...
-                    meta.update({file['key']: file['meta']})
+
+                # Update corresponding metadata
+                file_meta = file.get('meta', {})
+                if self._show_path:
+                    if 'data' in file_meta:
+                        file_meta['data']['image_path'] = file['path']
+                    else:
+                        file_meta['data'] = {'image_path': file['path']}
+                meta.update({file['key']: file_meta})
```
maingoh: nit: if you add only one key, is it not equivalent to `meta[file['key']] = file_meta` ?

PR #84, opened 2019-07-30
Support JSON string wildcard

leo-p:
As an output we already support two types of JSON:
- Classic JSON: `deepo ... -o prediction.json`
- Integer wildcard JSON: `deepo ... -o prediction_%04d.json` where the wildcard is replaced with the frame number.

This PR adds support for:
- String wildcard JSON: `deepo ... -o prediction_%s.json` where the wildcard is replaced with the frame name.


deepomatic/cli/output_data.py
```diff
@@ -248,18 +248,30 @@ def __init__(self, descriptor, **kwargs):
         super(JsonOutputData, self).__init__(descriptor, **kwargs)
         self._i = 0
         self._to_studio_format = kwargs.get('studio_format')
-        # Check if the output is a wild card or not
+        # Check if the output is a string wildcard
         try:
-            descriptor % self._i
+            descriptor % 'string'
+            self._wildcard_type = 'string'
```
maingoh: Can you use an enum ?

deepomatic/cli/output_data.py
```diff
@@ -285,7 +297,10 @@ def output_frame(self, frame):
                 self._all_predictions.append(predictions)
         # Otherwise we write them to file directly
         else:
-            json_path = os.path.splitext(self._descriptor % self._i)[0]
+            if self._wildcard_type == 'integer':
+                json_path = os.path.splitext(self._descriptor % self._i)[0]
+            elif self._wildcard_type == 'string':
+                json_path = os.path.splitext(self._descriptor % frame.name)[0]
```
maingoh: I forgot, what is `frame.name` ? Can it contains space and special chars ?

deepomatic/cli/output_data.py
```diff
@@ -248,18 +248,30 @@ def __init__(self, descriptor, **kwargs):
...
+            except TypeError:
+                self._wildcard_type = None
+                if self._to_studio_format:
+                    self._all_predictions = {'tags': [], 'images': []}
+                else:
+                    self._all_predictions = []
 
     def close(self):
         if self._all_predictions is not None:
+            # If we have only one element, make it a pure Vulcan JSON
```
maingoh: This is where things get unconsistent .. We shouldn't need magical behaviors.
Please keep always a list.

tests/test_infer.py
```diff
@@ -89,12 +97,12 @@ def test_e2e_json_infer(outputs, expected):
...
 
 
 def test_e2e_image_infer_json_threshold():
-    run_infer(INPUTS['IMAGE'], [OUTPUTS['JSON']], expect_nb_json=1, extra_opts=['-t', '0.5'])
+    run_infer(INPUTS['IMAGE'], [OUTPUTS['NO_WILDCARD_JSON']], expect_nb_json=1, extra_opts=['-t', '0.5'])
 
 
 def test_e2e_image_infer_json_studio():
-    run_infer(INPUTS['IMAGE'], [OUTPUTS['JSON']], expect_nb_json=1, studio_format=True, extra_opts=['--studio_format'])
+    run_infer(INPUTS['IMAGE'], [OUTPUTS['NO_WILDCARD_JSON']], expect_nb_json=1, studio_format=True, extra_opts=['--studio_format'])
```
maingoh: Note for one day: stronger tests, check filenames etc.

deepomatic/cli/output_data.py
leo-p: `frame.name` is composed of the file name without the extension plus the frame_number if it's a video plus the model used. For instance:
- `video.mp4` -> `video_000xx_fashion-v4`
- `img.png` -> `img_fashion-v4`

maingoh: Maybe normalize it without accent, spaces, lower cases etc ? Or you want it to be exactly the same ?

maingoh: Just curious, is there a specific need ? All filenames would be kind of the same ?
You would have `video_000xx_fashion-v4` for each frame instead of `video_000xx` (less duplication). Would it make more sense to organize this by directory `video/fashion-v4/0000x.jpg` ?

leo-p: I'm not sure we want to normalize it, the whole idea is to be able to very easily match filename and their corresponding prediction. Normalizing might cause issues if you have file `leo1.png` and `léo1.png`.
Also, the need is not that much when you have a single file (whether it be video or image), but instead when you have several files.

For instance:
```
images
├── img1.png
└── img2.png
```
with `deepo infer -i images -o predictions/pred_%s.json -r 123`, would yield something like:
```
predictions
├── pred_img1_123.json
└── pred_img2_123.json
```

tests/utils.py
```diff
@@ -63,7 +65,10 @@ def check_directory(directory,
                     assert 'tags' in data
                     assert 'images' in data
                 else:
-                    assert 'outputs' in data
+                    if isinstance(data, dict):
```
maingoh: I think it is not needed anymore

leo-p: Right now we store a list of Vulcan JSON only when we concatenate all predictions in one JSON.

Otherwise we store a pure Vulcan JSON (dict), see here: https://github.com/Deepomatic/deepocli/blob/ae59ad402667afa9e06c39a18abe39e51fecc0e7/deepomatic/cli/output_data.py#L289

You think we should store only Vulcan JSON lists?

maingoh: oh ok, then it's fine !

PR #85, opened 2019-07-30
Add resilience to image error

leo-p:
This PR make image error non-blocking to add resilience. For instance with 2 proper images and 1 corrupt images the result is:

```sh
$ deepo infer -i images -o img.json -r fashion-v4
[INFO 2019-07-30 16:57:31,260] Input processing:   0%|          | 0/3 [00:00<?, ?it/s]
[WARNING 2019-07-30 16:57:31,273] Cannot cast recognition ID into a number, trying with a public recognition model
libpng warning: iCCP: CRC error
libpng warning: iCCP: invalid distance too far back
libpng error: [00][02][10][80]: invalid chunk type
[ERROR 2019-07-30 16:57:31,276] Encountered an unexpected exception during handling of frame img_corrupted_fashion-v4, skipping it: OpenCV(3.4.3) /Users/travis/build/skvark/opencv-python/opencv/modules/imgcodecs/src/grfmt_base.cpp:145: error: (-10:Unknown error code -10) Raw image encoder error: Empty JPEG image (DNL not supported) in function 'throwOnEror'

[INFO 2019-07-30 16:57:32,879] Input processing:  33%|###3      | 1/3 [00:01<00:03,  1.62s/it]
[INFO 2019-07-30 16:57:33,136] Input processing:  67%|######6   | 2/3 [00:01<00:00,  1.07it/s]
[INFO 2019-07-30 16:57:33,192] Input processing: 100%|##########| 3/3 [00:01<00:00,  1.55it/s]
[WARNING 2019-07-30 16:57:33,192] Encountered an unexpected exception during handling of 1 frames out of 3
```

### DIFF
--- a/deepomatic/cli/cmds/infer.py
+++ b/deepomatic/cli/cmds/infer.py
@@ -138,13 +138,13 @@ class PrepareInferenceThread(thread_base.Thread):
     def process_msg(self, frame):
         try:
             _, buf = cv2.imencode('.jpg', frame.image)
-            buf_bytes = buf.tobytes()
-            frame.buf_bytes = buf_bytes
-            self.current_messages.add_frame(frame)
-            return frame
         except Exception as e:
             LOGGER.error('Could not decode image for frame {}: {}'.format(frame, e))
-        return None
+            return None
+        buf_bytes = buf.tobytes()
+        frame.buf_bytes = buf_bytes
+        self.current_messages.add_frame(frame)
+        return frame
 
 
 class SendInferenceGreenlet(thread_base.Greenlet):
--- a/deepomatic/cli/cmds/infer.py
+++ b/deepomatic/cli/cmds/infer.py
@@ -136,11 +136,15 @@ class BlurImagePostprocessing(object):
 
 class PrepareInferenceThread(thread_base.Thread):
     def process_msg(self, frame):
-        _, buf = cv2.imencode('.jpg', frame.image)
-        buf_bytes = buf.tobytes()
-        frame.buf_bytes = buf_bytes
-        self.current_messages.add_frame(frame)
-        return frame
+        try:
+            _, buf = cv2.imencode('.jpg', frame.image)
+            buf_bytes = buf.tobytes()
+            frame.buf_bytes = buf_bytes
+            self.current_messages.add_frame(frame)
+            return frame
+        except Exception as e:
+            LOGGER.error('Could not decode image for frame {}: {}'.format(frame, e))
+        return None
 
 
 class SendInferenceGreenlet(thread_base.Greenlet):
--- a/deepomatic/cli/thread_base.py
+++ b/deepomatic/cli/thread_base.py
@@ -158,15 +158,7 @@ class ThreadBase(object):
                             empty = True
 
                     if self.input_queue is None or not empty:
-                        try:
-                            msg_out = self.process_msg(msg_in)
-                        except Exception as e:
-                            msg_out = None
-                            if msg_in is None:
-                                LOGGER.error("Encountered an unexpected exception during routine: {}".format(e))
-                            else:
-                                LOGGER.error("Encountered an unexpected exception during handling of frame {}, skipping it: {}".format(msg_in.name, e))
-
+                        msg_out = self.process_msg(msg_in)
                         if msg_out is not None:
                             self.put_to_output(msg_out)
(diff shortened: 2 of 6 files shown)


deepomatic/cli/thread_base.py
```diff
@@ -157,8 +157,17 @@ def _run(self):
                             msg_in = self.pop_input()
                         except Empty:
                             empty = True
+
                     if self.input_queue is None or not empty:
-                        msg_out = self.process_msg(msg_in)
+                        try:
```
maingoh: I would prefer doing this explicitely where needed. Where you open the image, just warn and return None. Just like here https://github.com/Deepomatic/deepocli/blob/621d11d56638e272094ec1871725f901ed41f6ff/deepomatic/cli/cmds/infer.py#L194

maingoh: Now I am mixed, skipping a frame is something expected. Catching any exception is not expected and it should never happen.
So:
- Either we consider when it happen this is critical and exit, just like before
- Either we consider we can recover but this is unclear and might put the program in an unknown and buggy state (infinite loop), no CTRL+C working.

My preference would be to exit properly if something unexpected happen, and just add bug fixes/new catch for the ones we consider as expected and recoverable (just like skipping an image).

deepomatic/cli/cmds/infer.py
```diff
@@ -136,11 +136,15 @@ def __call__(self, frame):
...
-        frame.buf_bytes = buf_bytes
-        self.current_messages.add_frame(frame)
-        return frame
+        try:
+            _, buf = cv2.imencode('.jpg', frame.image)
+            buf_bytes = buf.tobytes()
+            frame.buf_bytes = buf_bytes
+            self.current_messages.add_frame(frame)
+            return frame
+        except Exception as e:
```
maingoh: Catch only what you need:
```python
try:
            _, buf = cv2.imencode('.jpg', frame.image)
except Exception as e:
            LOGGER.error('Could not decode image for frame {}: {}'.format(frame, e))
            self.current_messages.forget_frame(frame)
            return None

buf_bytes = buf.tobytes()
frame.buf_bytes = buf_bytes
self.current_messages.add_frame(frame)
return frame

```

deepomatic/cli/thread_base.py
```diff
@@ -299,7 +300,23 @@ def cleanup(self):
...
         if self.cleanup_func is not None:
             self.cleanup_func()
-        self.pbar.close()
+
+        # If needed, display the number of errors
+        total_inputs = self.pbar.total
+        inputs_without_error = self.pbar.n
+        if inputs_without_error < total_inputs:
+            self.pbar.update(total_inputs - inputs_without_error)
+            self.pbar.close()
```
maingoh: Remove the else close ?

deepomatic/cli/cmds/infer.py
leo-p: Yes to catch only what we need.

Fortunately, we don't need to `forget_frame()` because no message has been added yet. This is done with the last line before return `self.current_messages.add_frame(frame)`

alemoreau: the first output of cv2.imencode returns wheter the encoding was successful or not 
`retval, buf = cv2.imencode('.jpg', frame.image)`
Does it raise an exception when the image cannot be encoded ? Maybe you just need to check the retval value

deepomatic/cli/thread_base.py
leo-p: Added some details to explain the logic.

deepomatic/cli/cmds/infer.py
maingoh: > Fortunately, we don't need to forget_frame() because no message has been added yet. This is done with the last line before return self.current_messages.add_frame(frame)

True !

> the first output of cv2.imencode returns wheter the encoding was successful or not
retval, buf = cv2.imencode('.jpg', frame.image)

This would be better indeed, though it seems Leo had to try/catch to skip invalid images ?
Maybe the exception was actually raised after imencode ?

leo-p: Yes, the exception is indeed raised during `imencode`, the following code still crashes because `print(retval)` is never reached.

```
retval, buf = cv2.imencode('.jpg', frame.image)
print(retval)
```

PR #88, opened 2019-08-01
Expose greenlet number as environment variable

leo-p:
Expose a new environment variable `DEEPOMATIC_CLI_ADD_IMAGES_CONCURRENCY` which defaults to 5.

### DIFF
--- a/deepomatic/cli/cmds/feedback.py
+++ b/deepomatic/cli/cmds/feedback.py
@@ -14,7 +14,7 @@ from ..thread_base import Pool, MainLoop
 
 ###############################################################################
 
-GREENLET_NUMBER = int(os.getenv('DEEPOMATIC_CLI_ADD_IMAGES_GREENLET_NUMBER', 10))
+GREENLET_NUMBER = int(os.getenv('DEEPOMATIC_CLI_ADD_IMAGES_CONCURRENCY', 5))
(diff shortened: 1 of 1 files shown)


deepomatic/cli/cmds/feedback.py
```diff
@@ -14,7 +14,7 @@
 
 ###############################################################################
 
-GREENLET_NUMBER = 10
+GREENLET_NUMBER = int(os.getenv('DEEPOMATIC_CLI_ADD_IMAGES_GREENLET_NUMBER', 10))
```
thomas-riccardi: maybe instead name with `parallel` somewhere: greenlet number is really an internal implementation detail
Suggestion: `DEEPOMATIC_CLI_ADD_IMAGES_PARALLEL_LEVEL`

thomas-riccardi: maybe also change the default value to 1 for now?

maingoh: It's okay, it is a quite internal env var. I prefer `DEEPOMATIC_CLI_ADD_IMAGES_CONCURRENCY`.

maingoh: `1` will be really slow, I think 5 is a good default.

Issue #87, opened 2019-08-01
Reproduce directory structure with recursive option and string wildcard json


leo-p:
When using the recursive option with a json string wildcard, we should reproduce the input directory structure. Otherwise file can collide.

With directory:
```
dir
├── subdir1
│   └── img.png
└── subdir2
    └── img.png
```
And command:
```
deepo infer -i dir -R -o predictions/pred_%s.json -r 123
```
Will result in:
```
predictions
└── pred_img_123.json
```

leo-p:
We also need to deal with the case where we have `img.png` and `img.jpg`.

PR #90, opened 2019-08-10
Reproduce minimal input directory structure

leo-p:
Closes https://github.com/Deepomatic/deepocli/issues/87.

Now when using a JSON output with string wildcard `pred/%s.json` and the `--recursive` option, the minimal input directory structure is preserved to avoid collisions.

For instance:
```sh
$ tree images
images
├── img.png
└── subdir
    ├── img.png
    ├── subdir1
    │   └── img.png
    └── subdir2
        └── img.png
$ deepo infer -i images/ -o predictions/%s.json -R -r 123
$ tree predictions
predictions
├── img_123.json
└── subdir
    ├── img_123.json
    ├── subdir1
    │   └── img_123.json
    └── subdir2
        └── img_123.json

### DIFF
--- a/deepomatic/cli/output_data.py
+++ b/deepomatic/cli/output_data.py
@@ -255,12 +255,16 @@ class JsonOutputData(OutputData):
         super(JsonOutputData, self).__init__(descriptor, **kwargs)
         self._i = 0
         self._to_studio_format = kwargs.get('studio_format')
+        self._reproduce_input_dir_structure = False
 
         # Check if the output is a string wildcard
         try:
             descriptor % 'string'
             self._wildcard_type = WildCardType.STRING
             self._all_predictions = None
+            if kwargs.get('recursive', False):
+                self._reproduce_input_dir_structure = True
+                self._input_path = kwargs.get('input')
         except TypeError:
             # Check if the output is an integer wildcard
             try:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: deepomatic/cli/output_data.py [hunk 1]
```diff
@@ -255,12 +255,16 @@ def __init__(self, descriptor, **kwargs):
...
+        self._reproduce_input_dir_structure = False
 
         # Check if the output is a string wildcard
         try:
             descriptor % 'string'
             self._wildcard_type = WildCardType.STRING
             self._all_predictions = None
+            if kwargs.get('recursive', False):
+                self._reproduce_input_dir_structure = True
+                self._input_path = kwargs.get('input')
```
maingoh: this variable doesn't seem to be set if not recursive

deepomatic/cli/output_data.py
```diff
@@ -255,12 +255,16 @@ def __init__(self, descriptor, **kwargs):
         super(JsonOutputData, self).__init__(descriptor, **kwargs)
         self._i = 0
         self._to_studio_format = kwargs.get('studio_format')
+        self._reproduce_input_dir_structure = False
```
maingoh: Only nits: 
- Can't you set it here `self._reproduce_input_dir_structure = kwargs.get('recursive', False)`. 
- `preserve` instead of `reproduce`

deepomatic/cli/output_data.py
```diff
@@ -302,10 +306,25 @@ def output_frame(self, frame):
                 self._all_predictions.append(predictions)
         # Otherwise we write them to file directly
         else:
+            # Build the prediction json path
             if self._wildcard_type == WildCardType.INTEGER:
                 json_path = os.path.splitext(self._descriptor % self._i)[0]
             elif self._wildcard_type == WildCardType.STRING:
-                json_path = os.path.splitext(self._descriptor % frame.name)[0]
+                if not self._reproduce_input_dir_structure:
```
maingoh: I don't understand why we need to depend on this flag, can't we always preserve the same structure of the input ?

tests/test_infer.py
```diff
@@ -110,3 +134,7 @@ def test_e2e_image_infer_json_studio():
 
 def test_e2e_image_corrupted_infer_json():
     run_infer(INPUTS['IMAGE_CORRUPTED'], [OUTPUTS['INT_WILDCARD_JSON']], expect_nb_json=0)
+
+
+def test_e2e_image_infer_reproduce_input_dir_struct():
+    run_infer(INPUTS['DIRECTORY'], [OUTPUTS['STR_WILDCARD_JSON']], expect_nb_json=2, expect_nb_subdir=1, extra_opts=['--recursive'])
```
maingoh: Idea for later: having a `check_run_*_output_structure` for tests that have less than 5 frames:
```
    dir_struct = {
    "root": {
        "folder2": {
            "item2": None, 
            "item1": None
        }, 
        "folder1": {
            "subfolder1": {
                "item2": None, 
                "item1": None
            }, 
            "subfolder2": {
                "item3": None
            }
        }
    }
}
    check_run_infer_output_structure(INPUTS['DIRECTORY'], [OUTPUTS['STR_WILDCARD_JSON']], dir_struct, extra_opts=['--recursive'])

```

Might help http://code.activestate.com/recipes/577879-create-a-nested-dictionary-from-oswalk/

Or maybe using https://docs.python.org/3/library/filecmp.html#the-dircmp-class

deepomatic/cli/output_data.py
leo-p: I think it only makes sense to reproduce the input directory structure if the output is a string wildcard.
- If it's not a wildcard, only one JSON with all predictions will be outputted.
- If it's an integer wildcard, the JSON will be outputted sequentially `pred_1.json`, `pred_2.json` so I don't think it makes sense to divide them into directories.

Agreed for `_preserve_input_dir_structure`

deepomatic/cli/output_data.py
leo-p: See answer on line 258.

Issue #91, opened 2019-08-10
Manage extensions collision


leo-p:
Today we don't handle collisions due to extensions.

The two files `img.png` and `img.jpg` will result in the same prediction files, drawn images, JSON key, etc.

We might want to incorporate the extension in the `frame.name` to have a proper unique identifier.

PR #92, opened 2019-08-10
Add extension in frame name

leo-p:
Closes https://github.com/Deepomatic/deepocli/issues/91.

Adds extension in `frame.name` to avoid extension collision:
- Before two files `img.jpg` and `img.png` would get the same frame name `img_123`.
- Now the extension is taken into account resulting in `img_jpg_123` and `img_png_123`.

This also deals with `img.jpg` and `img.JPG` cases.

This breaks retrocompatibility with `--from_file` for offline predictions files but this feature has not been released on PyPi yet.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -148,7 +148,7 @@ def init_files_setup():
     # Download JSON files
     vulcan_json_pth = download(tmpdir, 'https://s3-eu-west-1.amazonaws.com/deepo-tests/vulcain/json/vulcan.json', 'vulcan.json')
     studio_json_pth = download(tmpdir, 'https://s3-eu-west-1.amazonaws.com/deepo-tests/vulcain/json/studio.json', 'studio.json')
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -148,7 +148,7 @@ def init_files_setup():
     # Download JSON files
     vulcan_json_pth = download(tmpdir, 'https://s3-eu-west-1.amazonaws.com/deepo-tests/vulcain/json/vulcan.json', 'vulcan.json')
     studio_json_pth = download(tmpdir, 'https://s3-eu-west-1.amazonaws.com/deepo-tests/vulcain/json/studio.json', 'studio.json')
-    offline_pred_pth = download(tmpdir, 'https://s3-eu-west-1.amazonaws.com/deepo-tests/vulcain/json/offline_pred.json', 'offline_pred.json')
+    offline_pred_pth = download(tmpdir, 'https://s3-eu-west-1.amazonaws.com/deepo-tests/vulcain/json/offline_pred2.json', 'offline_pred.json')
```
leo-p: @maingoh this is a kind of test because the `frame.name` is stored in the JSON. So if it doesn't match, the tests fail.

PR #93, opened 2019-08-12
Add amqp output

alemoreau:
AMQP output enables easier interconnection with other programs. For now, only direct exchanges are used, the format for specifying the routing key is `-o amqp.routing_key`

### DIFF
--- a/deepomatic/cli/output_data.py
+++ b/deepomatic/cli/output_data.py
@@ -52,7 +52,7 @@ def get_output(descriptor, kwargs):
             return JsonOutputData(descriptor, **kwargs)
         elif DirectoryOutputData.is_valid(descriptor):
             return DirectoryOutputData(descriptor, **kwargs)
-        elif descriptor == 'amqp':
+        elif AMQPOutputData.is_valid(descriptor):
             return AMQPOutputData(descriptor, **kwargs)
         elif descriptor == 'stdout':
             return StdOutputData(**kwargs)
@@ -164,11 +164,21 @@ class ImageOutputData(OutputData):
             write_frame_to_disk(frame, path)
 
 class AMQPOutputData(OutputData):
+    @classmethod
+    def is_valid(cls, descriptor):
+        return descriptor.startswith('amqp')
+
     def __init__(self, descriptor, **kwargs):
         super(AMQPOutputData, self).__init__(descriptor, **kwargs)
         if not RPC_PACKAGES_USABLE:
             raise DeepoRPCUnavailableError('RPC not available')
         self._amqp_url = kwargs['amqp_url']
+        try:
+            # descriptor is expected to have amqp.routing_key format
+            _, self._routing_key = descriptor.split('.', 1)
+        except ValueError:
+            self._routing_key = None
+
         self._client = None
         self._queue = None
 
@@ -183,7 +193,7 @@ class AMQPOutputData(OutputData):
         else:
             if self._client is None:
                 self._client = Client(self._amqp_url)
-                self._queue = self._client.new_queue()
+                self._queue = self._client.new_queue(self._routing_key)
                 LOGGER.info('Output queue: %s' % self._queue.routing_key)
             # using pickle for now, TODO: protobuf ?
             message = pickle.dumps({
--- a/deepomatic/cli/output_data.py
+++ b/deepomatic/cli/output_data.py
@@ -198,6 +198,7 @@ class AMQPOutputData(OutputData):
             # using pickle for now, TODO: protobuf ?
             message = pickle.dumps({
                 'frame': frame.image,
+                'buffer': frame.buf_bytes,
                 'predictions': frame.predictions,
                 'filename': frame.filename
             })
--- a/deepomatic/cli/output_data.py
+++ b/deepomatic/cli/output_data.py
@@ -5,11 +5,19 @@ import cv2
 import imutils
 import logging
 import traceback
+import signal
+import multiprocessing
+import pickle
 from .thread_base import Thread
 from .common import Empty, write_frame_to_disk, SUPPORTED_IMAGE_OUTPUT_FORMAT, SUPPORTED_VIDEO_OUTPUT_FORMAT
 from .cmds.studio_helpers.vulcan2studio import transform_json_from_vulcan_to_studio
 from .exceptions import DeepoUnknownOutputError, DeepoSaveJsonToFileError
 
+try:
+    from deepomatic.rpc.client import Client
+    RPC_PACKAGES_USABLE = True
+except ImportError:
+    RPC_PACKAGES_USABLE = False
 
 LOGGER = logging.getLogger(__name__)
 DEFAULT_OUTPUT_FPS = 25
@@ -44,6 +52,8 @@ def get_output(descriptor, kwargs):
             return JsonOutputData(descriptor, **kwargs)
         elif DirectoryOutputData.is_valid(descriptor):
             return DirectoryOutputData(descriptor, **kwargs)
+        elif descriptor == 'amqp':
+            return AMQPOutputData(descriptor, **kwargs)
         elif descriptor == 'stdout':
             return StdOutputData(**kwargs)
         elif descriptor == 'window':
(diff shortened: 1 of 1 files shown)


deepomatic/cli/output_data.py
```diff
@@ -153,6 +163,46 @@ def output_frame(self, frame):
...
+    def output_frame(self, frame):
+        if frame.output_image is None:
+            LOGGER.warning('No frame to output.')
+        else:
+            if self._client is None:
+                self._client = Client(self._amqp_url)
+                self._queue = self._client.new_queue(self._routing_key)
+                LOGGER.info('Output queue: %s' % self._queue.routing_key)
+            # using pickle for now, TODO: protobuf ?
+            message = pickle.dumps({
```
maingoh: I am shared, do you need all these variables (especially the opencv image) ? I would avoid using pickle as much as possible https://www.benfrederickson.com/dont-pickle-your-data/
Please use a protobuf message.

deepomatic/cli/output_data.py
```diff
@@ -153,6 +163,46 @@ def output_frame(self, frame):
...
+    def __init__(self, descriptor, **kwargs):
+        super(AMQPOutputData, self).__init__(descriptor, **kwargs)
+        if not RPC_PACKAGES_USABLE:
+            raise DeepoRPCUnavailableError('RPC not available')
+        self._amqp_url = kwargs['amqp_url']
+        try:
+            # descriptor is expected to have amqp.routing_key format
+            _, self._routing_key = descriptor.split('.', 1)
+        except ValueError:
+            self._routing_key = None
```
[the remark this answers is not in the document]
alemoreau: -o amqp would only create a tmp queue. I'm not sure if it's a good idea though, maybe the `amqp.routing_key` format should be enforced

deepomatic/cli/output_data.py
```diff
@@ -153,6 +163,46 @@ def output_frame(self, frame):
...
+    def close(self):
+        if self._client is not None:
+            self._client.remove_queue(self._queue)
+            self._client.amqp_client.ensured_connection.close()
+
+    def output_frame(self, frame):
+        if frame.output_image is None:
+            LOGGER.warning('No frame to output.')
+        else:
+            if self._client is None:
```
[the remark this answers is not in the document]
alemoreau: This class is instanciated in the main thread, but the client will be use in the output thread, so the client should be instanciated in the output thread, shoudn't it ?

deepomatic/cli/output_data.py
alemoreau: :+1: for protobuf

deepomatic/cli/output_data.py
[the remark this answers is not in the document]
alemoreau: I get `gevent.exceptions.ConcurrentObjectUseError: This socket is already used by another greenlet: <bound method Waiter.switch of <gevent.__waiter.Waiter object at 0x7f96dd7ad688>>` when instanciated in the __init__ method

deepomatic/cli/output_data.py
```diff
@@ -153,6 +163,46 @@ def output_frame(self, frame):
...
             write_frame_to_disk(frame, path)
 
+class AMQPOutputData(OutputData):
+    @classmethod
+    def is_valid(cls, descriptor):
+        return descriptor.startswith('amqp')
+
+    def __init__(self, descriptor, **kwargs):
+        super(AMQPOutputData, self).__init__(descriptor, **kwargs)
+        if not RPC_PACKAGES_USABLE:
```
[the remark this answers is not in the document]
alemoreau: https://github.com/Deepomatic/deepocli/pull/93/commits/68e8429a5adb71376072579ddc1febd5ca9ba046#diff-dd511fb275735fb69a76404698c1ca76R21

deepomatic/cli/output_data.py
```diff
@@ -153,6 +163,46 @@ def output_frame(self, frame):
         finally:
             write_frame_to_disk(frame, path)
 
+class AMQPOutputData(OutputData):
+    @classmethod
+    def is_valid(cls, descriptor):
+        return descriptor.startswith('amqp')
```
leo-p: Maybe check for `amqp://` to make sure it's not just a random file name?

deepomatic/cli/output_data.py
```diff
@@ -153,6 +163,46 @@ def output_frame(self, frame):
...
+            self._client.amqp_client.ensured_connection.close()
+
+    def output_frame(self, frame):
+        if frame.output_image is None:
+            LOGGER.warning('No frame to output.')
+        else:
+            if self._client is None:
+                self._client = Client(self._amqp_url)
+                self._queue = self._client.new_queue(self._routing_key)
+                LOGGER.info('Output queue: %s' % self._queue.routing_key)
```
leo-p: Maybe switch it to `LOGGER.debug`? The user already knows the output queue because he typed it in the CLI.

deepomatic/cli/workflow/rpc_workflow.py
```diff
@@ -1,33 +1,57 @@
...
+                from deepomatic.rpc.response import wait_responses
+                from deepomatic.rpc.helpers.v07_proto import create_recognition_command_mix
+                from deepomatic.rpc.helpers.proto import create_v07_images_command
+            except ImportError as e:
+                raise DeepoRPCUnavailableError('RPC version mismatch: %s' % e)
 
-class RpcRecognition(AbstractWorkflow):
+            try:
+                return old_new(cls, *args, **kwargs)
+            except:
```
leo-p: Maybe be more specific on the error type?

deepomatic/cli/workflow/rpc_workflow.py
```diff
@@ -1,33 +1,57 @@
...
+                return old_new(cls)
 
-    class InferResult(AbstractWorkflow.AbstractInferResult):
+        return __new__
+    cls.__new__ = override_new(cls.__new__)
+    return cls
 
+@requires_deepomatic_rpc
+class RpcRecognition(AbstractWorkflow):
+    @requires_deepomatic_rpc
```
leo-p: Isn't checking `RpcRecognition` for RPC enough? Do we also need to check `InferResult` as it's defined only in `RpcRecognition`?

alemoreau: This class could be instantiated and that would crash without a  DeepoRPCUnavailableError.
But I agree, since its only instanciated inside RpcRecognition instance, which can only be instanciated if deepo rpc is available, that should be enough

deepomatic/cli/output_data.py
alemoreau: The descriptor format is actually only amqp.routing_key.
But to prevent conflit with other formats, maybe the descriptor format could be amqp#routing_key or something like this

leo-p: Could we use the same as for the `-u` parameter? `-u amqp://user:password@localhost:port` ?

alemoreau: Mhh its not the same thing
amqp://user:password@localhost:port  is an URI (used to connect to the rabbitmq broker)
amqp#routing_key is just an internal format (used to create a queue in the amqp broker)

alemoreau: We need a way to attach a string parameter to the amqp output. It could be another param, but since it's only used for amqp output, it simpler like this (it's like specifying a path for the file outputs, except its prefixed with amqp. If it were amqp://routin_key, it would be ambiguous since its not an URI, I think)

deepomatic/cli/output_data.py
```diff
@@ -5,15 +5,25 @@
...
 
 LOGGER = logging.getLogger(__name__)
 DEFAULT_OUTPUT_FPS = 25
 
+try:
+    from deepomatic.rpc.client import Client
+    from deepomatic.rpc.buffers.protobuf.cli.Message_pb2 import Message
+    from google.protobuf.json_format import Parse
+except:
+    pass
```
maingoh: I think it would be preferable to do:

```
if RPC_IS_AVAILABLE:
    from deepomatic.rpc.client import Client
    from deepomatic.rpc.buffers.protobuf.cli.Message_pb2 import Message
    from google.protobuf.json_format import Parse
```

(no silent exception)

deepomatic/cli/output_data.py
```diff
@@ -153,6 +172,72 @@ def output_frame(self, frame):
...
+        if self._client is not None:
+            self._client.remove_queue(self._queue)
+            self._client.amqp_client.ensured_connection.close()
+
+    def output_frame(self, frame):
+        if frame.output_image is None:
+            LOGGER.warning('No frame to output.')
+        else:
+            message = Message()
+            Parse(json.dumps({
```
maingoh: Use https://github.com/protocolbuffers/protobuf/blob/c132a4aa165d8ce2b65af62d4bde4a7ce08d07c3/python/google/protobuf/json_format.py#L418

maingoh: This way you can directly set the source ;)

deepomatic/cli/output_data.py
```diff
@@ -153,6 +172,72 @@ def output_frame(self, frame):
...
+                            'inputs': [{
+                                'image': {
+                                    'source': ''
+                                }
+                            }]
+                        }
+                    }
+                }
+            }), message)
+            message.command.input_mix.v07_inputs.inputs[0].image.source = frame.buf_bytes
```
maingoh: Keep the prefix `BINARY_IMAGE_PREFIX`

deepomatic/cli/workflow/rpc_workflow.py
```diff
@@ -4,30 +4,53 @@
...
     from deepomatic.rpc.client import Client
-    RPC_PACKAGES_USABLE = True
-except ImportError:
-    RPC_PACKAGES_USABLE = False
-
-# If the deepomatic-rpc module is installed, then we try to import the other modules. An error here might indicate a
-# version mismatch, in which case we want to know which one and why, i.e. we don't want to catch the error but we want
-# to display it to the end user.
-if RPC_PACKAGES_USABLE:
     from deepomatic.rpc import v07_ImageInput, BINARY_IMAGE_PREFIX
```
maingoh: Do you still need this if it is in `requires_deepomatic_rpc` ?

deepomatic/cli/output_data.py
```diff
@@ -5,15 +5,25 @@
 import imutils
 import logging
 import traceback
+import signal
```
maingoh: I think you can remove those 3 imports

leo-p:
Should we merge this PR or is it still WIP?

PR #98, opened 2019-08-29
Display both score and label by default

leo-p:
Previously, if you didn't specify any of the `-L/--draw_labels` or `-S/--draw_scores` option with the `draw` command, nothing was displayed. Which is at odd with the definition of the `draw` command.

Now the default behaviour is to display both. We also introduce two new options `--no_draw_labels` and `--no_draw_scores` which are mutually exclusive with their counterpart.

This is retro compatible with the previous options.

### DIFF
--- a/deepomatic/cli/cli_parser.py
+++ b/deepomatic/cli/cli_parser.py
@@ -64,11 +64,14 @@ def argparser_init():
         parser.add_argument('--from_file', type=str, dest='pred_from_file', help="Uses prediction from a Vulcan or Studio JSON.")
 
     score_group = draw_parser.add_mutually_exclusive_group()
-    score_group.add_argument('-S', '--draw_scores', help="Overlay the prediction scores. Default behavior.", action="store_true")
-    score_group.add_argument('--no_draw_scores', help="Do not overlay the prediction scores.", action="store_false")
+    score_group.add_argument('-S', '--draw_scores', dest='draw_scores', help="Overlay the prediction scores. Default behavior.", action="store_true")
+    score_group.add_argument('--no_draw_scores', dest='draw_scores', help="Do not overlay the prediction scores.", action="store_false")
+    score_group.set_defaults(draw_scores=True)
+
     label_group = draw_parser.add_mutually_exclusive_group()
-    label_group.add_argument('-L', '--draw_labels', help="Overlay the prediction labels. Default behavior.", action="store_true")
-    label_group.add_argument('--no_draw_labels', help="Do not overlay the prediction labels.", action="store_false")
+    label_group.add_argument('-L', '--draw_labels', dest='draw_labels', help="Overlay the prediction labels. Default behavior.", action="store_true")
+    label_group.add_argument('--no_draw_labels', dest='draw_labels', help="Do not overlay the prediction labels.", action="store_false")
+    label_group.set_defaults(draw_labels=True)
 
     blur_parser.add_argument('-M', '--blur_method', help="Blur method to apply, either 'pixel', 'gaussian' or 'black', defaults to 'pixel'.", default='pixel', choices=['pixel', 'gaussian', 'black'])
     blur_parser.add_argument('-B', '--blur_strength', help="Blur strength, defaults to 10.", default=10)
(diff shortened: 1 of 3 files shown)


maingoh:
> Now if you don't specify any option, it is as if you specified -LS/--draw_labels --draw_scores, both are displayed.

This looks like a weird behavior no ? I would suggest, by default display everything and provide flags to disable some of them `--no-draw-labels/--no-draw-scores`.

leo-p:
> This looks like a weird behavior no ? I would suggest, by default display everything and provide flags to disable some of them `--no-draw-labels/--no-draw-scores`.

Option A: The way it's currently implemented has the quality of being consistent with the previous behaviour, meaning the command you were using before this PR is still working after this PR.

Option B: Your solution is indeed more clear, if we are to change the options completely we could go with two mutually exclusive options `--draw_labels_only` and `--draw-scores_only`. (It doesn't make sense to use `deepo draw` without drawing either the labels or score, this is a way to enforce it).

Which one makes more sense to you?

maingoh:
Agreed, but I am not sure in both case that those options are used a lot for now ? That's why I thought we could break compatibility. For option B, I think it can make sense to draw more things maybe ? Frame number, timestamp, tracking id etc. Having two mutually options would be restrictive for the future, no ?

deepomatic/cli/cli_parser.py
```diff
@@ -63,8 +63,12 @@ def argparser_init():
         parser.add_argument('-F', '--fullscreen', help="Fullscreen if window output.", action="store_true")
         parser.add_argument('--from_file', type=str, dest='pred_from_file', help="Uses prediction from a Vulcan or Studio JSON.")
 
-    draw_parser.add_argument('-S', '--draw_scores', help="Overlays the prediction scores.", action="store_true")
-    draw_parser.add_argument('-L', '--draw_labels', help="Overlays the prediction labels.", action="store_true")
+    score_group = draw_parser.add_mutually_exclusive_group()
```
maingoh: What is the difference with ?
```
parser.add_argument('--draw_scores', dest='draw_scores', action='store_true')
parser.add_argument('--no_draw_scores', dest='draw_scores', action='store_false')
parser.set_defaults(draw_scores=True)
```

deepomatic/cli/cmds/infer.py
```diff
@@ -32,8 +32,8 @@ def get_coordinates_from_roi(roi, width, height):
 class DrawImagePostprocessing(object):
 
     def __init__(self, **kwargs):
-        self._draw_labels = kwargs.get('draw_labels', False)
-        self._draw_scores = kwargs.get('draw_scores', False)
+        self._draw_labels = kwargs.get('draw_labels', False) or kwargs.get('no_draw_labels', True)
```
maingoh: All defaults should be defined in the argparse (as early as possible)

deepomatic/cli/cmds/infer.py
```diff
@@ -44,12 +44,15 @@ def __call__(self, frame):
         for pred in frame.predictions['outputs'][0]['labels']['predicted']:
             # Build legend
             label = u''
-            if self._draw_labels:
-                label = pred['label_name']
+            label_label = pred['label_name']
```
maingoh: What is the difference with previously ?

deepomatic/cli/cmds/infer.py
```diff
@@ -83,8 +86,9 @@ def __call__(self, frame):
                 text_corner = substract_tuple(text_corner, (x_offset, y_offset))
 
                 # Finally draw everything
-                cv2.rectangle(output_image, background_corner1, background_corner2, BACKGROUND_COLOR, -1)
-                cv2.putText(output_image, label, text_corner, cv2.FONT_HERSHEY_SIMPLEX, FONT_SCALE, TEXT_COLOR, 1)
+                if label != '':
+                    cv2.rectangle(output_image, background_corner1, background_corner2, BACKGROUND_COLOR, -1)
```
maingoh: It's a small rectangle to attach the label with the box right ?
Why calculating background_corner1 and so if we don't draw them (the if should probably be earlier). Also I guess it depends only on `self.draw_labels` no ?

leo-p: Agreed for the `if` that should be earlier to avoid useless computation.

As a whole it depends on both `self.draw_labels` and `self.draw_scores` plus maybe other variables in the future. `label` already encompasses all those checks.

deepomatic/cli/cmds/infer.py
leo-p: Nothing, it's a relic of the previous temporary implementation.

deepomatic/cli/cli_parser.py
leo-p: The mutually exclusive group ensures it's not possible to do `deepo draw ... --draw_scores --no_draw_scores`.

PR #114, opened 2019-09-25
More resilient to inference errors

### LINKED ISSUE #75: Inference: retry on API/network errors
Related to https://github.com/Deepomatic/deepomatic-client-python/issues/17

Currently we are not resilient to unexpected bad status code 5xx and the whole program will quit if it happens.

### LINKED ISSUE #102: Warning at the end is confusing
We should display a clearer reporting at the end: 
```
Exiting: errors={} processed={} total={}
```

We should'nt rely on the progress bar to get the number of errors, we should count them. 

https://github.com/Deepomatic/deepocli/blob/a8ee864ae6708c33d52216c0aa5a549c25c06146/deepomatic/cli/thread_base.py#L315

maingoh:
Close #102 
Close #75 

- Use `http_retry` fonctionnality from deepomatic-api 0.9.0 to retry on http/network errors
- Skip inference errors and count them

### DIFF
--- a/deepomatic/cli/thread_base.py
+++ b/deepomatic/cli/thread_base.py
@@ -74,7 +74,7 @@ class CurrentMessages(object):
     def report_success(self):
         self.report_successes(1)
 
-    def report_successes(self, nb_successes)
+    def report_successes(self, nb_successes):
         with self.lock():
             self.nb_success += nb_successes
 
@@ -333,7 +333,7 @@ class MainLoop(object):
                           self.current_messages.nb_errors -
                           self.current_messages.nb_successes)
 
-        LOGGER.info('Exiting ..')
+        LOGGER.info('Exiting..')
         LOGGER.info('Summary: errors={} uncompleted={} successful={} total={}.'.format(self.current_messages.nb_errors,
                                                                                        nb_uncompleted,
                                                                                        self.current_messages.nb_successes,
--- a/deepomatic/cli/thread_base.py
+++ b/deepomatic/cli/thread_base.py
@@ -48,6 +48,8 @@ class CurrentMessages(object):
         self.heap_lock = Lock()
         self.messages = []
         self.nb_errors = 0
+        self.nb_successes = 0
+        self.nb_added_messages = 0
 
     def lock(self):
         return blocking_lock(self.heap_lock)
@@ -55,6 +57,7 @@ class CurrentMessages(object):
     def add_message(self, msg):
         with self.lock():
             heapq.heappush(self.messages, msg)
+            self.nb_added_messages += 1
 
     def get_min(self):
         with self.lock():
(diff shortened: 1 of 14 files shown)


deepomatic/cli/thread_base.py
```diff
@@ -304,21 +318,12 @@ def cleanup(self):
...
-
-        # Display errors or images stopped if needed
-        if inputs_without_error < total_inputs and self.stop_asked:
-            LOGGER.warning('Handled {} frames out of {} before stopping.'.format(inputs_without_error, total_inputs))
-        elif inputs_without_error < total_inputs:
-            LOGGER.warning('Encountered an unexpected exception during handling of {} frames out of {}.'.format(
-                total_inputs - inputs_without_error, total_inputs
-            ))
+        LOGGER.info('Exiting: errors={} successful={} total={}'.format(self.current_messages.nb_errors,
+                                                                       self.pbar.n - self.current_messages.nb_errors,
```
alemoreau: It is a little weird to use the progressbar to get the successful count. (What if we change the progressbar implementation in the future ?) Maybe add a nb_success variable to current_messages, just like nb_errors ?

maingoh: Yes I could do this ! (got a bit lazy)

maingoh: Done

deepomatic/cli/thread_base.py
```diff
@@ -304,21 +328,18 @@ def cleanup(self):
...
+                          self.current_messages.nb_successes)
 
-        # Display errors or images stopped if needed
-        if inputs_without_error < total_inputs and self.stop_asked:
-            LOGGER.warning('Handled {} frames out of {} before stopping.'.format(inputs_without_error, total_inputs))
-        elif inputs_without_error < total_inputs:
-            LOGGER.warning('Encountered an unexpected exception during handling of {} frames out of {}.'.format(
-                total_inputs - inputs_without_error, total_inputs
-            ))
+        LOGGER.info('Exiting..')
```
leo-p: Is this information useful if we provide the summary afterwards?

deepomatic/cli/thread_base.py
```diff
@@ -304,21 +328,18 @@ def cleanup(self):
...
+        LOGGER.info('Summary: errors={} uncompleted={} successful={} total={}.'.format(self.current_messages.nb_errors,
```
leo-p: Do we need to provide the full summary when we have been only successful?

maingoh: Well this is a question I asked myself, but I finally think we should be consistent. 
As this is a summary, the user might expect to see it. If it is not there, it might be considered as a bug (program did not exit properly ?).

deepomatic/cli/thread_base.py
maingoh: True, removed

deepomatic/cli/thread_base.py
leo-p: True, let's go for consistency!

PR #94, opened 2019-08-12
Add workflow (pipeline) support

alemoreau:
For some use cases, we need to combine multiple inference, like a pipeline.
This PR enables to define a pipeline as a json file, are executing it using `-f pipeline.json`
Later (or in this PR) we could add more stream operations like `filter, crop, ` to create more complex pipelines, as well as intermediary outputs.

PR #132, opened 2019-10-23
Fix inititialization of pools + allow to reuse pipeline

maingoh:
Extracted from #94 
- Threads must be initialized before starting any pool or it can lead to concurrent errors
- We can now reuse the pipeline to add other pools at the end (see `OutputThread`). Improved `task_done()` to handle specific scenarios. Also now task_done is automatically called.

### DIFF
--- a/deepomatic/cli/output_data.py
+++ b/deepomatic/cli/output_data.py
@@ -58,7 +58,13 @@ def get_outputs(descriptors, kwargs):
     return [get_output(descriptor, kwargs) for descriptor in descriptors]
 
 
+class NotProcessedYet(object):
+    pass
+
+
 class OutputThread(Thread):
+    NOT_PROCESSED_YET = NotProcessedYet()
+
     def __init__(self, exit_event, input_queue, output_queue, current_messages,
                  on_progress, postprocessing, **kwargs):
         super(OutputThread, self).__init__(exit_event, input_queue,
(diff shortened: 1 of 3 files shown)


deepomatic/cli/output_data.py
```diff
@@ -102,10 +108,31 @@ def pop_input(self):
             frame = super(OutputThread, self).pop_input()
         return frame
 
+    def put_to_output(self, frame_out):
+        if frame_out is self.NOT_PROCESSED_YET:
```
leo-p: I can't find where `self.NOT_PROCESSET_YET` is defined.

deepomatic/cli/output_data.py
```diff
@@ -58,7 +58,13 @@ def get_outputs(descriptors, kwargs):
     return [get_output(descriptor, kwargs) for descriptor in descriptors]
 
 
+class NotProcessedYet(object):
```
leo-p: Maybe explain its function?

deepomatic/cli/output_data.py
maingoh: It's a static attribute https://github.com/Deepomatic/deepocli/pull/132/files#diff-c47c647771bc242f6f2743d2cba9e7d3R66

deepomatic/cli/output_data.py
maingoh: It's a class, I added some comments :)

deepomatic/cli/thread_base.py
```diff
@@ -166,13 +166,12 @@ def put_to_output(self, msg_out):
...
             try:
                 self.output_queue.put(msg_out, block=False)
-                self.task_done()
                 break
             except Full:
                 # don't touch until we have non performance regression tests
                 gevent.sleep(SLEEP_TIME)
 
-    def task_done(self):
+    def task_done(self, msg_in, msg_out):
```
alemoreau: Why do we need two args ?

deepomatic/cli/output_data.py
```diff
@@ -102,10 +113,31 @@ def pop_input(self):
...
         return frame
 
+    def put_to_output(self, frame_out):
+        if frame_out is self.NOT_PROCESSED_YET:
+            # Nothing to output, the frame has not been processed yet
+            return
+        super(OutputThread, self).put_to_output(frame_out)
+
+    def task_done(self, frame_in, frame_out):
+        if frame_out is self.NOT_PROCESSED_YET:
```
alemoreau: Should the NOT_PROCESSED_YET logic belong to the thread_base instead ? Other phases of the pipeline could benefit from it ?

deepomatic/cli/thread_base.py
maingoh: I think it provides more flexibility to the thread that overrides the method (msg_in might be useful)

deepomatic/cli/output_data.py
maingoh: Yes, it could but since it's a quite specific behavior, I would prefer not to pollute thread_base for now. We can move it if one day if it is needed.

Issue #136, opened 2019-11-12
Implement communication with the worker-nn by using the Workflow command


vdel:
Currently, we need to fill in the recognition_id in the operate show case UI. This was required because we use the "Recognition" RPC command, which requires this number.

See the number `40030` in this screen shot:
<img width="491" alt="Capture d’écran 2019-11-12 à 15 21 47" src="https://user-images.githubusercontent.com/3261014/68679690-a4e9b380-0560-11ea-8ce7-aaa5c14a6a77.png">

However, the spirit of what we are building is to rely on Workflows: they take no argument in the RPC command (not `show_discarded`, not `max_num_results` --> this should be a parameter of the Recognition node of the workflow).

We thus need to switch on the "Workflow" RPC command, which will have the benefit of removing the "Recognition" field in the UI.

## How to fix ?

As soon as the workflows are released, we probably just need to switch to `create_workflow_command_mix` here: https://github.com/Deepomatic/deepocli/blob/b17c743dff6aaaf7f278b605a3da514e34c43da9/deepomatic/cli/workflow/rpc_workflow.py#L89

PR #137, opened 2019-11-12
Workflow command mix

alemoreau:
Use workflow command mix when recognition version id is not provided as argument
fix https://github.com/Deepomatic/deepocli/issues/136

### DIFF
--- a/deepomatic/cli/cli_parser.py
+++ b/deepomatic/cli/cli_parser.py
@@ -108,7 +108,7 @@ def argparser_init():
     # Define model group for infer draw blur
     for cmd in ['infer', 'draw', 'blur']:
         group = model_groups[cmd]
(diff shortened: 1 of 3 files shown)


deepomatic/cli/cli_parser.py
```diff
@@ -108,7 +108,7 @@ def argparser_init():
     # Define model group for infer draw blur
     for cmd in ['infer', 'draw', 'blur']:
         group = model_groups[cmd]
-        group.add_argument('-r', '--recognition_id', required=True, help="Neural network recognition version ID.")
+        group.add_argument('-r', '--recognition_id', help="Neural network recognition version ID.")
```
maingoh: we could force type int (not related to this PR though)

PR #147, opened 2019-12-03
Only use Queue class, remove LifoQueue

alemoreau:
LifoQueue was for the old pipelining, now it introduces a frame leak


deepomatic/cli/input_data.py
```diff
@@ -120,16 +120,13 @@ def input_loop(kwargs, postprocessing=None):
         LOGGER.error(str(e))
         sys.exit(1)
 
-    # For realtime, queue should be LIFO
-    # TODO: might need to rethink the whole pipeling for infinite streams
     # IMPORTANT: maxsize is important, it allows to regulate the pipeline and avoid to pushes too many requests to rabbitmq when we are already waiting for many results
```
leo-p: Nit: move this comment above `queues` definition?

PR #149, opened 2020-01-07
Fix print json validation error

### LINKED ISSUE #141: Studio: wrong summary uncompleted when adding images
Airtable url: [https://api.airtable.com/v0/appgWOl8ecHRBewEz/Bugs%20%26%20Issues/rec6lnTlkcztCtiHR](https://airtable.com/tblL2bTpvN3KcH8BB/viwxchJz20QpypLBP/rec6lnTlkcztCtiHR)
When I upload image through deepocli, I get a wrong summary in the end. See the last line.

```
[INFO 2019-11-26 09:34:38,494] Uploading images:  98%|#########8| 1750/1782 [11:41<00:12,  2.49it/s]
[INFO 2019-11-26 09:34:40,599] Uploading images:  99%|#########8| 1760/1782 [11:43<00:08,  2.50it/s]
[INFO 2019-11-26 09:34:41,914] Uploading images:  99%|#########9| 1770/1782 [11:45<00:04,  2.51it/s]
[INFO 2019-11-26 09:34:50,608] Uploading images: 100%|##########| 1782/1782 [11:53<00:00,  2.50it/s]
[INFO 2019-11-26 09:34:50,661] Summary: errors=0 uncompleted=-1782 successful=1782 total=1782.
```

romyli:
Close #141

Prints the error message and the instance concerned if the JSON is not validated in the add_images command. For example:
`Error with Vulcan JSON : 123 is not of type 'string' in the instance [0, 'location']`
`Error with Studio JSON : '0.548' is not of type 'number' in ['images', 0, 'annotated_regions', 0, 'region', 'ymin']`

Issue #141: Temporary fix for the negative value of uncompleted image count in the summary (`Summary: errors=0 uncompleted=-1 successful=1 total=1.`)


deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -6,7 +6,7 @@
 from .vulcan2studio import transform_json_from_vulcan_to_studio
 from ...thread_base import Greenlet
 from ...common import SUPPORTED_IMAGE_INPUT_FORMAT, SUPPORTED_VIDEO_INPUT_FORMAT
-from ...json_schema import is_valid_studio_json, is_valid_vulcan_json
+from ...json_schema import is_valid_studio_json, is_valid_vulcan_json, validate_json
```
leo-p: Are `is_valid_studio_json` and `is_valid_vulcan_json` still used? If not do we need them in `json_schema.py`?

```suggestion
from ...json_schema import validate_json
```

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -115,14 +115,20 @@ def fill_queue(self, files, dataset_name, commit_pk):
...
+                    # If it's a Studio json, use it directly
+                    if schema_type == 'Studio':
+                        studio_json = json_data
+                        LOGGER.warning("{} JSON {} validated".format(schema_type, upload_file))
+                    # If it's a Vulcan json, transform it to Studio
+                    elif schema_type == 'Vulcan':
+                        studio_json = transform_json_from_vulcan_to_studio(json_data)
+                        LOGGER.warning("Vulcan JSON {} validated and transformed to Studio format".format(upload_file))
+                # If the JSON is not valid, print the error
+                elif error is not None:
```
leo-p: Can we merge both `elif error is not None:` and `else:` statements? Is it possible to have an unvalidated json without an error?

deepomatic/cli/json_schema.py
```diff
@@ -112,3 +112,28 @@ def is_valid_studio_json(json_data):
...
+def validate_json(json_data):
+    """
+    Validate a JSON using the Studio and Vulcan schema
+    Returns:
+    - is_valid: True if the JSON is valid
+    - error: ValidationError raised if not valid
+    - schema_type: Studio or Vulcan, or None if both schema raise an error at the root of the JSON
+    """
+    is_valid, error, schema_type = False, None, None
+    schema_dict = {'Studio':STUDIO_JSON_SCHEMA, 'Vulcan':VULCAN_JSON_SCHEMA}
```
leo-p:
```suggestion
    schema_dict = {'Studio': STUDIO_JSON_SCHEMA, 'Vulcan': VULCAN_JSON_SCHEMA}
```

deepomatic/cli/json_schema.py
```diff
@@ -112,3 +112,28 @@ def is_valid_studio_json(json_data):
...
+    for schema_name, json_schema in schema_dict.items():
+        try:
+            validate(instance=json_data, schema=json_schema)
+            is_valid = True
+            schema_type = schema_name
+            break
+        except Exception as e:
```
leo-p: Is it possible to catch directly the specific JSON error?

deepomatic/cli/cmds/studio_helpers/file.py
romyli: It is not possible to have an unvalidated json without an error, but I wasn't sure how to determine the desired json type (vulcan/studio) if it raises an error with both schemas. If the error is raised at the root of the JSON, `validate_json` returns no error so that we raise DeepoUploadJsonError without detailing the error as the type of json is not clear

deepomatic/cli/json_schema.py
romyli: You mean `from jsonschema import ValidationError` and catch it with `except ValidationError as e:` ?

leo-p: For instance, yes! It would make sure we don't silence other exceptions by error.

deepomatic/cli/json_schema.py
```diff
@@ -112,3 +112,28 @@ def is_valid_studio_json(json_data):
...
+    schema_dict = {'Studio': STUDIO_JSON_SCHEMA, 'Vulcan': VULCAN_JSON_SCHEMA}
+    for schema_name, json_schema in schema_dict.items():
+        try:
+            validate(instance=json_data, schema=json_schema)
+            is_valid = True
+            schema_type = schema_name
+            break
+        except Exception as e:
+            # If the error did not happen at the root, return the error and the current schema type
+            if len(e.absolute_path) > 0:
```
maingoh: Why is it necessary ?

deepomatic/cli/json_schema.py
```diff
@@ -112,3 +112,28 @@ def is_valid_studio_json(json_data):
...
+                error = e
+                schema_type = schema_name
+                break
+    return is_valid, error, schema_type
```
maingoh: Couldn't we just return `error` and `schema_type` ? We could consider that if `error` is None then it is valid.

deepomatic/cli/json_schema.py
romyli: First, we check if the json is a Studio JSON, and if the validation error is at the root of the json, it means it may be a Vulcan JSON. Thus, we don't return an error, we break and check if the Vulcan schema type matches

deepomatic/cli/json_schema.py
romyli: If we have `is_valid_json=False` and `error=None` it means we have not identified which type of JSON schema it is. This is used only in the error message printed in `file.py`, if it is not necessary to detail this I can remove it

deepomatic/cli/json_schema.py
```diff
@@ -106,9 +142,37 @@ def is_valid_json_with_schema(json_data, json_schema):
 
 def is_valid_studio_json(json_data):
```
leo-p: `is_valid_studio_json()` and `is_valid_vulcan_json()` don't seem to be used anymore. Maybe remove them?

deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -115,14 +117,20 @@ def fill_queue(self, files, dataset_name, commit_pk):
...
+                    if schema_type == JSONSchemaType.STUDIO:
+                        studio_json = json_data
+                        LOGGER.warning("{} JSON {} validated".format(schema_type, upload_file))
+                    # If it's a Vulcan json, transform it to Studio
+                    elif schema_type == JSONSchemaType.VULCAN:
+                        studio_json = transform_json_from_vulcan_to_studio(json_data)
+                        LOGGER.warning("Vulcan JSON {} validated and transformed to Studio format".format(upload_file))
+                # If the JSON is not valid, print the error
+                elif error is not None:
+                    LOGGER.warning("Error with {} JSON : {} in the instance {}".format(schema_type, error.message, list(error.path)))
```
leo-p: What if we have an error with `schema_type=None`, maybe we could make the warning message clearer?

deepomatic/cli/json_schema.py
maingoh: Okay understood, my idea was to reuse the error variable: `error = e.message` for json error and `error = 'Unknown json format'` when the format is not identified. But anyway that is okay like this for me :)

deepomatic/cli/json_schema.py
maingoh: Okay, sounds reasonnable :)

deepomatic/cli/thread_base.py
```diff
@@ -85,6 +85,13 @@ def report_errors(self, nb_errors):
         with self.lock():
             self.nb_errors += nb_errors
 
+    def report_message(self):
```
maingoh: Does it works as expected ? I think you can add "Close #141" in your PR description

deepomatic/cli/cmds/studio_helpers/file.py
romyli: I have changed the `elif` condition to make it clearer: `elif schema_type is not None and error is not None:`

PR #153, opened 2020-02-03
Let commands declare their parser

### LINKED ISSUE #152: Let each command parse the command line
`cli_parser.py` is getting too complicated. We should be able to refactor it so that we just have:
```
inference.parse_cmd_line(subparsers)
feedback.parse_cmd_line(subparsers)
cameras.parse_cmd_line(subparsers)
```
And move each command parsing in the command file. This probably imply to put `noop` `infer` `draw` and `blur` as subcommand instead of root commands (breaking change).

maingoh:
Close #152 

- Moved parsers from blur, draw, infer, noop into `cmds/infer.py`.
- Renamed `cmds/feedback.py` into `cmds/studio.py`.
- Moved parsers from add_images `cmds/studio.py`.
- Few details to make the code more readable.


deepomatic/cli/cmds/infer.py
```diff
@@ -206,3 +220,184 @@ def process_msg(self, frame):
...
+    # threads, and the exit code is 0.
+    # If SIGINT or SIGTERM is sent again during this shutdown phase, the threads are killed, and the exit code is 2.
+    if exit_event.is_set():
+        sys.exit(1)
+    elif stop_asked >= 2:
+        sys.exit(2)
+
+
+def setup_cmd_line_parsers(inference_parsers):
+    inference_parsers['infer'].set_defaults(func=input_loop)
```
leo-p: Maybe make sure that the dict `keys` already exist?

maingoh: I think it doesn't bring much, since if it doesn't exist it will be a programming error and will be raised right away with a `KeyError: infer`.

maingoh:
Yes this is probably doable and better, but imply a bit more changes I think. 
I will create an issue :)

Issue #146, opened 2019-12-03
Allow to upload to studio programmatically


maingoh:
We should be able to programmatically upload on studio. 

Multiple solutions:
-  Allow to disable gevent monkey patch and import studio helpers from deepomatic-cli. This is a bit hacky and implies to import deepomatic-cli which is not really done for this (quite short term).
- Move studio helpers in a standalone repo and package and reuse deepomatic-client-python as dependency.
- Move studio helpers in deepomatic-client-python (probably best option) but this imply changing again the name of the python package which is `deepomatic-api` (it won't be only the api). We can argue this will be for all deepomatic APIs, or we can also rename it `deepomatic-http` or `deepomatic-http-client`.

PR #155, opened 2020-02-05
Camera commands

maingoh:


deepomatic/cli/__init__.py
```diff
@@ -1,5 +1,8 @@
+import os
 from gevent.monkey import patch_all
-patch_all(thread=False, time=False)  # NOQA
+
+if os.getenv('DEEPOMATIC_CLI_GEVENT_MONKEY_PATCH', '1') == '1':
+    patch_all(thread=False, time=False)  # NOQA
```
alemoreau: In what situation shouldn't we monkey patch ?

deepomatic/cli/cmds/camera.py
```diff
@@ -0,0 +1,228 @@
...
+        # Output json ?
+        pass
+
+
+def draw(args):
+    rpc, protobuf = import_rpc_package(should_raise=True)
+    name = str(uuid.uuid4())
+    with temp_camera(name, args.camera_address):
+        # TODO: draw_camera_results() # With json ?
+        pass
```
alemoreau: The existing pipeline can be used for that (using noop command and amqp input)

deepomatic/cli/cmds/camera.py
```diff
@@ -0,0 +1,228 @@
...
+    for subcmd in ['start', 'stop', 'delete']:
+        help_msg = '{} a camera.'.format(subcmd.capitalize())
+        add_parser(camera_subparser, parsers, subcmd, help_msg)
+
+    # Stream
+    add_parser(camera_subparser, parsers, 'list', 'List cameras.', add_name=False)
+    help_msg = ('Add and start a camera then exit.'
+                'The camera will continue to transmit frames until stopped by user (for production).')
+    add_parser(camera_subparser, parsers, 'stream', help_msg)
+
```
alemoreau: Preview command would be nice too (not 100% sure how to do that though)

deepomatic/cli/__init__.py
maingoh: Well, it is not so much related to this PR but address the first point of https://github.com/Deepomatic/deepocli/issues/146
(Since we don't have a pip package for the studio client and camera client, we could import `deepomatic.cli` without wanting getenv)

deepomatic/cli/cmds/camera.py
maingoh: I think it will be a bit different since we don't read anything, we just retrieve the results. Not sure the inference pipeline will help much (but I will see what I can do).

deepomatic/cli/cmds/camera.py
maingoh: I think I will just support the draw with output window.

PR #165, opened 2020-08-06
Fixing setup.py pip issue

ibreschi:


setup.py
```diff
@@ -23,6 +23,11 @@
 
 # Read requirements
 install_reqs = parse_requirements(os.path.join(here, 'requirements.txt'), session='hack')
+# for pip 20.2.1
+try:
+    requirements = [str(ir.req) for ir in install_reqs]
+except Exception:
+    requirements = [str(ir.requirement) for ir in install_reqs]
```
thomas-riccardi, line 30:
```suggestion
try:
    requirements = [str(ir.requirement) for ir in install_reqs]
    
except Exception:  # for pip < 20.1
    requirements = [str(ir.req) for ir in install_reqs]
```

it has possibly be broken by https://github.com/pypa/pip/pull/7747, shipped in pip 20.1

let's put the modern way first, more future proof

setup.py
```diff
@@ -23,6 +23,11 @@
 
 # Read requirements
 install_reqs = parse_requirements(os.path.join(here, 'requirements.txt'), session='hack')
+try:
+    requirements = [str(ir.requirement) for ir in install_reqs]
+
```
thomas-riccardi, line 28: nit: probably remove this useless empty line, not sure how it got there (I'm still not familiar with the multi line suggestion feature in github)..

maingoh:
I tend to remove parse_requirement fonction and just use sth line
```
with open('requirements.txt') as f:
    requirements = list(f.readlines())
```

I think it does basically the same thing and don't rely on api changes. You can remove empty lines eventually and maybe comments (i think they ignored anyway by requires param of setup)

thomas-riccardi:
> I tend to remove parse_requirement fonction and just use sth like
> 
> ```
> with open('requirements.txt') as f:
>     requirements = list(f.readlines())
> ```
> 
> I think it does basically the same thing and don't rely on api changes. You can remove empty lines eventually and maybe comments (i think they ignored anyway by requires param of setup)

with dependency pinning we do have comments everywhere in that file; if it's supported by `setuptools.setup()` then perfect; otherwise we will have to rollback to the previous version at some point...

maingoh:
It seems to work https://github.com/Deepomatic/deepomatic-rpc/blob/master/python/requirements.txt
Didn't test with empty line though

maingoh:
And https://github.com/Deepomatic/deepomatic-rpc/blob/cfa9634c76422abb506cf7eddefde2539b223467/python/setup.py#L12

setup.py
```diff
@@ -22,7 +17,8 @@
 os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
```
Zanoellia, line 17: Maybe it can be simplified with `pathlib`? I think they are equivalent 
```suggestion
from pathlib import Path
os.chdir(Path(__file__).parent.absolute())))
```

Zanoellia: Oh sorry I didn't see it was not part of your changes :sweat_smile:

maingoh:
As an example, people tend to avoid using pip internal now https://github.com/google/cloud-forensics-utils/commit/8e7721a162dd31f18775e0e5f9d59262fcfe9050

ibreschi:
> LGTM, you can add empty lines to the requirements.txt if you are not confident with my suggestion 😄 (the comments seems to work)

It seems to work even with empty lines.

PR #167, opened 2020-08-17
Platform tests

### LINKED ISSUE #166: Tests for new Platform commands
add tests to check the correct behavior of Platform commands.

ibreschi:
close #166


deepomatic/cli/lib/platform.py
```diff
@@ -8,10 +8,20 @@
 from deepomatic.api.http_helper import HTTPHelper
 
 
+def badstatus_catcher(func):
+    def func_wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except BadStatus as e:
+            print(f'Failed to run {func} {e}')
```
ibreschi, line 16: @maingoh do you know if I can print the func name  here?

tests/test_platform.py
```diff
@@ -0,0 +1,106 @@
...
+deploy_env = os.environ.copy()
+deploy_env['DEEPOMATIC_API_KEY'] = api_key
+deploy_env['DEEPOMATIC_API_URL'] = api_url
+
+
+def call_deepo(args):
+    args = args.split()
+    command = ['deepo'] + args
+    try:
+        res = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True, env=deploy_env)
```
maingoh, line 21: As we discussed I find using our `run_cmd` easier to debug and more straightforward. You can also test the a return value.
Using subprocess you don't really know what happened when the test fail and you need to parse stdout. 
You also need to install `deepo` before testing everything. Ideally you could test before installing.

tests/test_platform.py
```diff
@@ -0,0 +1,106 @@
...
+        assert res.returncode == 0
+        assert res.stderr == ''
+        return res.stdout.strip()
+    except AttributeError:
+        # For python 2.7
+        res = subprocess.check_output(command, universal_newlines=True, env=deploy_env)
```
maingoh, line 27: Can't you use check_output for both ?

tests/test_platform.py
```diff
@@ -0,0 +1,106 @@
...
+        return res.strip()
+
+
+@contextmanager
+def app():
+    args = "platform app create -n test -d abc -w {} -c {}".format(WORKFLOW_PATH, CUSTOM_NODES_PATH)
+    result = call_deepo(args)
+    _, app_id = result.split(':')
```
maingoh, line 35: parsing the output is a bit brittle no ? do we want to be that strict ? what if another print with `:` is added ?

tests/test_platform.py
```diff
@@ -0,0 +1,106 @@
...
+    yield app_id
+
+    args = "platform app delete --id {}".format(app_id)
+    result = call_deepo(args)
+
+
+@contextmanager
+def app_version():
+    with app() as app_id:
+        args = "platform appversion create -n test_av -d abc -a {} -r 61307 61306".format(app_id)
```
maingoh, line 46: `app-version` maybe

tests/test_platform.py
```diff
@@ -0,0 +1,106 @@
...
+        assert message == 'New app created with id'
+
+        args = "platform app update --id {} -d ciao".format(app_id)
+        result = call_deepo(args)
+        message = result.strip()
+        assert message == 'App{} updated'.format(app_id)
+
+        args = "platform app delete --id {}".format(app_id)
+        result = call_deepo(args)
+        message = result.strip()
```
maingoh, line 71: maybe `call_deepo` could strip the message

tests/test_platform.py
ibreschi: As discussed, I had a working run_cmd, but was forced to reuse the previous subprocess implementation. 
The reason behind it is that we can not change the environment of one already started process. 
As you can see I'm passing a deploy_env to the spawned process in which I change the DEEPOMATIC_API_KEY and DEEPOMATIC_API_URL env vars. 
Either we have a single API KEY or we are forced to do something like this.

tests/test_platform.py
ibreschi: yes but I loose the      
```
assert res.returncode == 0
assert res.stderr == ''
```

tests/test_platform.py
ibreschi: it already does, removing this line

tests/test_platform.py
ibreschi: I can improve the result parsing, but I have not many other alternatives. 
Anyway if a print with `:` is added, this will break the tests.

tests/test_platform.py
ibreschi: When we will remove python 2.7 we can remove this lines

tests/test_platform.py
ibreschi: I think that the command name is generated automatically from the class name in deepomatic/cli/cmds/utils.py :
```
self.name = self.__class__.__name__.lower().replace("command", "")
```
So given that a class name can not contain a `-` we can not have  `platform app-version`. 
But if you want we can have `app_version` 
Is it ok for you?

deepomatic/cli/lib/platform.py
```diff
@@ -8,10 +10,24 @@
 from deepomatic.api.http_helper import HTTPHelper
 
 
+LOGGER = logging.getLogger(__name__)
+
+
+def badstatus_catcher(func):
```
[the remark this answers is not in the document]
ibreschi: you are right

deepomatic/cli/lib/platform.py
```diff
@@ -20,33 +36,29 @@ def create_site(self, name, description, app_version_id):
...
-        files = {}
         with open(workflow_path, 'r') as f:
-            files['workflow_yaml'] = f
             workflow = yaml.safe_load(f)
 
-        if custom_nodes_path is not None:
-            files['custom_nodes_py'] = open(custom_nodes_path, 'r')
-
         # create using workflow server
         app_specs = [{
```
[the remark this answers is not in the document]
ibreschi: I'm not against parsing the workflow in the api. As we discussed once it will be cool to have a schema file that is used both in the API and in the WS to check the correctness of the workflow.yaml. 
Not having much time I think it will be done in future.

deepomatic/cli/lib/platform.py
```diff
@@ -57,15 +69,32 @@ def create_app(self, name, description, workflow_path, custom_nodes_path):
...
-    def update_app(self, app_id):
-        NotImplementedError()
+        ret = self.client.post('/apps-workflow', data=data_app, files=files, content_type='multipart/mixed')
+        return "New app created with id: {}".format(ret['app_id'])
 
+    @badstatus_catcher
+    def update_app(self, app_id, name, description):
+        data = {}
+
+        if name is not None:
```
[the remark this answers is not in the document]
ibreschi: Next time hugo ^^

deepomatic/cli/lib/platform.py
```diff
@@ -116,4 +155,7 @@ def upload(self):
 
     def validate(self):
         # TODO: implement
-        return True
+        if True:
```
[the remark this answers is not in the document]
ibreschi: Do you want I remove this command?

deepomatic/cli/lib/platform.py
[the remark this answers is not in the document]
maingoh: Agreed, just raising the concern, can be changed later

deepomatic/cli/lib/platform.py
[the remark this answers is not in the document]
maingoh: I would if it is not useful and we don't know what to put inside

deepomatic/cli/lib/platform.py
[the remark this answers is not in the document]
maingoh: :cry:

tests/test_platform.py
maingoh: With a fixture you could have temporarily changed the env var for the tests that needed them.
It's ok, it does what we want anyway, just maybe in the future we could uniformize with other tests.

tests/test_platform.py
maingoh: hmm I see, what advantages bring `subprocess.run` over the compatible `subprocess.call` or `subprocess.Popen` ?

tests/test_platform.py
maingoh: Using the same interpreter you would directly have the returned value no need to parse it (though you are not sure it is printed)

tests/test_platform.py
maingoh: `_` is fine for me, at least it is more readable and user friendly.
Or maybe we could replace the CamelCase to `-` separated ?
https://stackoverflow.com/questions/29916065/how-to-do-camelcase-split-in-python
then lower and join by `-`

deepomatic/cli/lib/platform.py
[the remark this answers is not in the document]
maingoh: ahah my point is that this decorator is not needed

deepomatic/cli/lib/platform.py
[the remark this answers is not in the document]
ibreschi: Removed the validate command

deepomatic/cli/lib/platform.py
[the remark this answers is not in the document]
ibreschi: removed the statush catcher

tests/test_platform.py
ibreschi: run is recommended for python 3.* : 
https://docs.python.org/3/library/subprocess.html

tests/test_platform.py
ibreschi: done

tests/test_platform.py
ibreschi: removed subprocess and using a contextmanager

tests/test_platform.py
ibreschi: improved a little the check but keeping like it is ^^

deepomatic/cli/lib/platform.py
```diff
@@ -70,14 +54,14 @@ def create_app(self, name, description, workflow_path, custom_nodes_path):
         if description is not None:
             data_app['desc'] = description
 
-        files = {'workflow_yaml': open(workflow_path, 'r')}
-        if custom_nodes_path is not None:
-            files['custom_nodes_py'] = open(custom_nodes_path, 'r')
+        with open(workflow_path, 'r') as w:
+            files = {'workflow_yaml': w}
+            if custom_nodes_path is not None:
+                files['custom_nodes_py'] = open(custom_nodes_path, 'r')
```
maingoh, line 60: nit: use with also

tests/utils.py
```diff
@@ -193,3 +193,41 @@ def run_cmd(cmds, inp, outputs, *args, **kwargs):
...
+    :param update: Dictionary of environment variables and values to add/update.
+
+    Usage:
+        a = os.environ.get('DEEPOMATIC_API_KEY')
+        print(a)
+        with modified_environ(DEEPOMATIC_API_KEY='24134'):
+            a = os.environ.get('DEEPOMATIC_API_KEY')
+            print(a)
+    """
+    env = os.environ
```
maingoh, line 216: ahaha nice ! I didn't ask that much !

maingoh (approved):
Congratulation ! It will be a major release in history :)

alemoreau (approved):

PR #172, opened 2020-10-23
Adapt import images requests for views in studio

qtreh:
- Add org argument for cmd add_images
- Stop using commit for studio requests

Tested with a local studio web endpoint
=> Small changes in vesta needed for it to work: https://github.com/Deepomatic/vesta/pull/3084

### DIFF
--- a/deepomatic/cli/cmds/studio.py
+++ b/deepomatic/cli/cmds/studio.py
@@ -71,6 +71,7 @@ def main(args):
 
     # Retrieve arguments
     dataset_name = args.get('dataset')
+    org_slug = args.get('org')
 
     paths = args.get('input', [])
     json_file = args.get('json_file', False)
@@ -86,7 +87,7 @@ def main(args):
     queue = Queue()
 
     dataset_files = DatasetFiles(clt.http_helper, queue)
-    total_files = dataset_files.post_files(dataset_name, files)
+    total_files = dataset_files.post_files(org_slug, dataset_name, files)
 
     exit_event = threading.Event()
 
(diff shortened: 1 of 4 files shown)


deepomatic/cli/cmds/studio_helpers/file.py
```diff
@@ -153,11 +153,11 @@ def fill_queue(self, files, dataset_name, commit_pk):
         self.flush_batch(url, batch)
         return total_files
 
-    def post_files(self, dataset_name, files):
+    def post_files(self, org_slug, dataset_name, files):
         # Retrieve endpoint
         try:
-            ret = self._helper.get('datasets/' + dataset_name + '/')
+            request = 'orgs/{}/projects/{}/'.format(org_slug, dataset_name)
```
maingoh, line 159: Is the org mandatory now @Alexis-Jacob ?

maingoh (commented):
Does the upload format change as well ?
The tests don't seem to succeed

deepomatic/cli/cmds/studio_helpers/file.py
Alexis-Jacob: yes, the org is the root of all api calls now

qtreh:
> Does the upload format change as well ?
> The tests don't seem to succeed

@maingoh The error is: resource = 'https://studio.deepomatic.com/api/orgs/travis-deepocli/projects/deepocli-feedback-test-detection/' is not found. The old API (current studio) works with /datasets/, the new one (staging, dev) with /projects/.

 Jenkins uses studio which has not been switched to the new API yet, we need to wait for a release. Once the release is done, this test should work, and we'll be able to merge.

deepomatic/cli/cmds/studio.py
```diff
@@ -128,6 +129,7 @@ def setup_cmd_line_subparser(studio_subparser):
     # Define studio group for add_images
     group = add_images_parser.add_argument_group('studio arguments')
     group.add_argument('-d', '--dataset', required=True, help="Deepomatic Studio dataset name.", type=str)
```
maingoh, line 131: Since the compatiblity is broken, maybe rename `dataset` into `project` ?

maingoh (approved):
LGTM

Alexis-Jacob (approved):

PR #174, opened 2020-10-31
add absolute frame index in frame metadata (index in input stream)

### LINKED ISSUE #157: Expose absolute index position of the frame
Considering the video doesn't contain duplicated frame as declared in #96, if we use an `--input_fps` or any option that skip frames, we would like to know the absolute position of the frame in the video. 
This involve incrementing a counter everytime we grab a frame (even if we don't decode it). 
Then we could expose it as metadata and/or in the framename.

alemoreau:
fix #157


deepomatic/cli/frame.py
```diff
@@ -2,21 +2,22 @@
...
 
 class Frame(object):
-    def __init__(self, name, filename, image, video_frame_index=None):
+    def __init__(self, name, filename, image, video_frame_index=None, absolute_video_frame_index=None):
         # The Frame object is used as a data exchanged in the different queues
         self.name = name  # name of the frame
         self.filename = filename  # the original filename from which the frame was extracted
         self.image = image  # an opencv loaded image (numpy array)
-        self.video_frame_index = video_frame_index  # index of the frame in the video sequence
+        self.video_frame_index = video_frame_index  # index of the frame in the output video sequence
```
maingoh, line 10: probably we should have a different comment for `video_frame_index` and `absolute_video_frame_index`.
`video_frame_index` is kind of the `decoded_frame_index` right (only the one that we decode) ?

deepomatic/cli/input_data.py
```diff
@@ -186,14 +189,15 @@ def _decode_next(self):
...
-            return Frame(self._name % self._i, self._filename, frame, self._i)
+            self._output_index += 1
+            return Frame(self._name % self._output_index, self._filename, frame, self._output_index, self._input_index)
 
     def _read_next(self):
         read, frame = self._cap.read()
         if read:
-            self._i += 1
-            return Frame(self._name % self._i, self._filename, frame, self._i)
+            self._input_index += 1
```
maingoh, line 198: Maybe this function could use `_grab_next` and `_decode_next` to avoid duplication of code in the future (don't change now)

deepomatic/cli/input_data.py
```diff
@@ -166,7 +167,8 @@ def _open_video(self, raise_exc=True):
 
     def __iter__(self):
         self._open_video()
-        self._i = 0
+        self._input_index = 0
```
maingoh, line 170: I am not sure to understand the naming of `_input_index` and `_output_index`. Maybe name the same than in Frame constructor ?

Zanoellia (approved):
Good to go!

I agree with @maingoh I was a bit disturbed by `input`/`output` naming even if it's generic.

deepomatic/cli/input_data.py
alemoreau: So from the input stream, some frames are selected (by skipping some input frames) to be put in the output.
What you called absolute index is actually the index from the input stream, while the 'normal' index is actually the index of the frame in the output.
I'm ok to put back 'absolute_index and index', or put a comment explaining those new names.
I find 'absolute index' quite clear, it is just that in the case of a stream, it may also refer to the index of the frame since the stream started, or if we provide an option to start reading a video in the middle for instance, it may refer to the index from the beginning.

deepomatic/cli/frame.py
```diff
@@ -2,21 +2,27 @@
...
```
maingoh, line 10: Maybe use `decoded_video_frame_index` as well here.
```suggestion
        self.decoded_video_frame_index = decoded_video_frame_index  # index from the decoded frames (skipped frames are not included)
```

deepomatic/cli/frame.py
```diff
@@ -2,21 +2,27 @@
...
+        self.absolute_video_frame_index = absolute_video_frame_index  # index of the frame in the input video sequence
```
maingoh, line 11:
```suggestion
        self.absolute_video_frame_index = absolute_video_frame_index  # absolute index of the frame in the video sequence (skipped frames included)
```

deepomatic/cli/frame.py
```diff
@@ -2,21 +2,27 @@
...
         self.buf_bytes = None
 
     def __str__(self):
-        return '<Frame name={} filename={} frame_number={} video_frame_index={} inference_async_result={}>'.format(
-            self.name, self.filename, self.frame_number, self.video_frame_index, self.inference_async_result)
+        return "<Frame {}>".format(' '.join("{}={}".format(key, getattr(self, key)) for key in [
+            'name',
+            'filename',
+            'frame_number',
+            'video_frame_index',
```
maingoh, line 23:
```suggestion
            'decoded_video_frame_index',
```

maingoh (approved):
Few suggestions for consistency and goo !

deepomatic/cli/input_data.py
```diff
@@ -172,6 +174,7 @@ def _stop_video(self, raise_exc=True):
 
     def _grab_next(self):
         grabbed = self._cap.grab()
+        self._absolute_video_frame_index += 1
```
maingoh, line 177: We should probably only increment if grabbed

PR #175, opened 2020-11-04
Earlier monkey patch

maingoh:
Also allow to disable gevent when not needed using `DEEPOMATIC_CLI_GEVENT_MONKEY_PATCH=1`


maingoh:
Should not be needed

Issue #185, opened 2020-12-21
Fix and improve camera commands


maingoh:
- `status` command doesn't exists in manager (use stream-info ? or circus status ?)
- rename `remove` command by `delete` to be consistent with other commands
- add other commands ? 
- put outside of `current` directory ? `deepo site camera` should be enough (current could avoid to provide site id though).
- add tests

PR #186, opened 2020-12-22
Bump deepomatic api

maingoh:
- Adapt to new deepomatic-api (for camera commands)
- Fix tests (API is now stricter)

### DIFF
--- a/deepomatic/cli/lib/camera.py
+++ b/deepomatic/cli/lib/camera.py
@@ -34,7 +34,7 @@ class Client(HTTPHelper):
         self.verify_ssl = verify_ssl
         self.host = host
 
-    def _setup_credentials(self):
+    def _setup_credentials(self, app_id, api_key):
         # No credentials
         pass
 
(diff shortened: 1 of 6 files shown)


deepomatic/cli/lib/camera.py
```diff
@@ -34,7 +34,7 @@ def _setup_host(self, host, verify_ssl):
         self.verify_ssl = verify_ssl
         self.host = host
 
-    def _setup_credentials(self):
+    def _setup_credentials(self, app_id, api_key):
```
Alexis-Jacob, line 37: not sure to understand the code here

maingoh: Yes, basically here we override this function from deepomatic-api because we don't have any auth for the camera server. 
The signature changed in deepomatic-api, that's why i adapted it.
In any case I think we will need to work back on it later as I am not totally sure everything works (we don't have tests), see https://github.com/Deepomatic/deepocli/issues/185

I am just making it "not crash" when running only `deepo -h`

Alexis-Jacob (approved):
lgtm

alemoreau (approved):

PR #197, opened 2021-03-04
Optional field id in format to upload images

Alexis-Jacob:
This PR gives the possibility to the user to upload a file without any id (or id == null)

### DIFF
--- a/deepomatic/cli/json_schema.py
+++ b/deepomatic/cli/json_schema.py
@@ -161,6 +161,9 @@ class JSONSchemaType:
         ],
         "additionalProperties": False,
         "properties": {
+            "id": {
+                "type": ["number", "null"]
+            },
             "data": {
                 "type": "array",
                 "items": {


maingoh (approved):
It would be nice to check why the test is passing with an invalid file, probably because of the `logger.error`.

Alexis-Jacob:
> It would be nice to check why the test is passing with an invalid file, probably because of the `logger.error`.

what it's strange it's that the data are uploaded ...

maingoh:
So I tested running only the test with the json using `pytest -k test_2e2_studio_json_upload
` and the data is not uploaded without this fix (count stays the same), as expected. Probably you felt that the data was uploaded because you were launching all studio tests which upload to the same project.
So we can merge !

Alexis-Jacob:
> So I tested running only the test with the json using `pytest -k test_2e2_studio_json_upload ` and the data is not uploaded without this fix (count stays the same), as expected. Probably you felt that the data was uploaded because you were launching all studio tests which upload to the same project.
> So we can merge !
Damn my bad ! thanks for looking for into it

PR #198, opened 2021-03-12
Raise warning not an error on unsupported files

antbertrand:
Added a little condition so that it ignores not supported files, warns the user about it, but doesn't raise any error.
On mac we have those .DS_Store files in every folder that made it difficult to upload local images using deepocli.

### DIFF
--- a/deepomatic/cli/json_schema.py
+++ b/deepomatic/cli/json_schema.py
@@ -230,7 +230,3 @@ def validate_json(json_data):
     return is_valid, error, schema_type
 
 
-if __name__ == "__main__":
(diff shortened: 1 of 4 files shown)


deepomatic/cli/json_schema.py
```diff
@@ -228,3 +228,9 @@ def validate_json(json_data):
                 schema_type = schema_name
                 break
     return is_valid, error, schema_type
+
+
+if __name__ == "__main__":
```
Alexis-Jacob, line 233: to remove I guess ?

Alexis-Jacob (approved):
please try to add a small test to cover this new condition

deepomatic/cli/json_schema.py
antbertrand: oh, yes my bad

deepomatic/cli/json_schema.py
```diff
@@ -228,3 +228,5 @@ def validate_json(json_data):
                 schema_type = schema_name
                 break
     return is_valid, error, schema_type
+
```
maingoh, line 231: nit: ideally don't add unecessary lines, especially if you don't touch the rest of the file :smile:

maingoh (approved):
Nice feature !

Issue #110, opened 2019-09-18
Draw font size


leo-p:
When using `deepo draw` it would be great if we could control the font size with an option or at least make it proportional to the image.

maingoh:
closed by #199

Issue #109, opened 2019-09-18
Draw background color for tags


leo-p:
When using `deepo draw` with a tagging/classificaiton model, all tags are displayed on a red background. It would be great if we could use a simple red/green/orange color code depending on the score and threshold.

maingoh:
closed by #199

PR #199, opened 2021-03-23
Expose font scale and thickness, and set background color depending on score

### LINKED ISSUE #190: Deepo draw label: add font size parameter and set color depending on score
- Add a parameter to set the font size (it is too small)
- Set the label color was depending on the score to visualize quickly the labels predicted (green if the score is high, orange if medium and red if low for example)

romyli:
1. Fix error if the output extension is in capital letter, and raise an error if the extension is not supported (otherwise we have the error "local variable fourcc referenced before assignment"
2. Expose font scale and font thickness to the user for `deepo draw`
3. Set the background color according to the score : if the threshold is none, it's green/orange/red if the score is >0.66, in between, or < 0.33 (otherwise the interval [threshold-1] is divided in 3)
4. Add parameter "unique_background_color" in case we want to disable the point 3 and have all the background in red as before

Closes #190 #110 #109

### DIFF
--- a/deepomatic/cli/cmds/utils.py
+++ b/deepomatic/cli/cmds/utils.py
@@ -62,6 +62,7 @@ class BuildDict(argparse.Action):
     """
     This class is used in argparse. It will transform a chain of name:values into a dict.
     """
+
     def __init__(self, option_strings, dest, nargs=None, **kwargs):
         self._nargs = nargs
         super(BuildDict, self).__init__(option_strings, dest, nargs=nargs, **kwargs)
@@ -149,7 +150,8 @@ def setup_model_cmd_line_parser(mode, cmd, inference_parsers):
         group.add_argument('-ft', '--font_thickness', dest='font_thickness', type=int,
                            help="Text font thickness, must be an int and defaults to 1",
                            default=1)
-        group.add_argument('--unique_background_color', dest='no_background_color', help="If true, draws labels with a unique red background color. By default, the background is red/orange/green depending on the threshold set.", action="store_true")
+        group.add_argument('--unique_background_color', dest='no_background_color',
+                           help="If true, draws labels with a unique red background color. By default, the background is red/orange/green depending on the threshold set.", action="store_true")
         score_group = group.add_mutually_exclusive_group()
         score_group.add_argument('-S', '--draw_scores', dest='draw_scores', help="Overlay the prediction scores. Default behavior.",
                                  action="store_true")
--- a/deepomatic/cli/cmds/utils.py
+++ b/deepomatic/cli/cmds/utils.py
@@ -143,6 +143,12 @@ def setup_model_cmd_line_parser(mode, cmd, inference_parsers):
     if cmd == "draw":
         # Define draw specific options
         group = inference_parsers.add_argument_group('drawing arguments')
+        group.add_argument('-fs', '--font_scale', dest='font_scale', type=float,
+                           help="Text font scale, defaults to 0.5",
+                           default=0.5)
+        group.add_argument('-ft', '--font_thickness', dest='font_thickness', type=int,
+                           help="Text font thickness, must be an int and defaults to 1",
+                           default=1)
         score_group = group.add_mutually_exclusive_group()
         score_group.add_argument('-S', '--draw_scores', dest='draw_scores', help="Overlay the prediction scores. Default behavior.",
                                  action="store_true")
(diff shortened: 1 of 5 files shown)


deepomatic/cli/cmds/utils.py
```diff
@@ -143,6 +144,14 @@ def setup_model_cmd_line_parser(mode, cmd, inference_parsers):
     if cmd == "draw":
         # Define draw specific options
         group = inference_parsers.add_argument_group('drawing arguments')
+        group.add_argument('-fs', '--font_scale', dest='font_scale', type=float,
+                           help="Text font scale, defaults to 0.5",
+                           default=0.5)
+        group.add_argument('-ft', '--font_thickness', dest='font_thickness', type=int,
+                           help="Text font thickness, must be an int and defaults to 1",
+                           default=1)
+        group.add_argument('--unique_background_color', dest='no_background_color',
```
maingoh, line 153: Only background color is a bit ambiguous, I suggest to give the ability to set the color using:
```
group.add_argument('--font_bg_color', default=None, nargs=3, help="Expect a B G R value. If set, draws labels with a unique background color. By default, the background is red/orange/green depending on the threshold set.", action="store_true")
```

deepomatic/cli/lib/inference.py
```diff
@@ -25,6 +25,9 @@
 FONT_SCALE = 0.5                        # Size of the font we draw in the image_output
 BOX_COLOR = (255, 0, 0)                 # Bounding box color (BGR)
 BACKGROUND_COLOR = (0, 0, 255)          # Text background color (BGR)
```
maingoh, line 27: I think you can remove this variable if you have a `--font_bg_color`

deepomatic/cli/lib/inference.py
```diff
@@ -69,7 +76,20 @@ def __call__(self, frame):
...
-            ret, baseline = cv2.getTextSize(label, cv2.FONT_HERSHEY_SIMPLEX, FONT_SCALE, 1)
+            ret, baseline = cv2.getTextSize(label, cv2.FONT_HERSHEY_SIMPLEX, self._font_scale, self._font_thickness)
+
+            score = pred.get('score')
+            if self._no_background_color:
+                background_color = BACKGROUND_COLOR
+            # Get background color depending on score
+            else:
+                interval = (1 - self._threshold) / 3
+                if score < interval:
```
maingoh, line 87: potentially the score can be None if `pred.get('score')` returns None which will make this comparison crash. 
So either you are confident (I think it should always be set) and thus use `score = pred['score']`, this way the error will be clearer if it fails. Otherwise just set a default if it is not there.

deepomatic/cli/lib/inference.py
```diff
@@ -69,7 +76,20 @@ def __call__(self, frame):
...
+                    background_color = DARK_RED_COLOR
+                elif score > 1 - interval:
```
maingoh, line 89: An idea if you want a smooth gradient from red to green https://stackoverflow.com/questions/340209/generate-colors-between-red-and-green-for-a-power-meter

deepomatic/cli/output_data.py
```diff
@@ -199,11 +199,13 @@ def is_valid(cls, descriptor):
 
     def __init__(self, descriptor, **kwargs):
         super(VideoOutputData, self).__init__(descriptor, **kwargs)
-        ext = os.path.splitext(descriptor)[1]
+        ext = os.path.splitext(descriptor)[1].lower()
```
maingoh, line 202: Nice !

deepomatic/cli/output_data.py
```diff
@@ -199,11 +199,13 @@ def is_valid(cls, descriptor):
...
         if ext == '.avi':
             fourcc = cv2.VideoWriter_fourcc('X', 'V', 'I', 'D')
         elif ext == '.mp4':
             fourcc = cv2.VideoWriter_fourcc('m', 'p', '4', 'v')
+        else:
+            raise DeepoUnknownOutputError('Unsupported video output extension')
```
maingoh, line 208: maybe add the extension in the error message

maingoh (commented):
Nice work !! Can you also add a small test for the non default way, which means when we set the text background color (no need to check the image, just make sure it does not crash)

deepomatic/cli/lib/inference.py
romyli: Ok, I will change it for  `score = pred['score']` I think it is better to get an error if there is no score (and it is what is currently used in the `draw_scores` part)

maingoh (approved):
So nice !! Congratulation :)

alemoreau (approved):
Great job !

Zanoellia (approved):
:sunglasses: :rainbow:

PR #200, opened 2021-03-30
Make tests fail on error logs

maingoh:
Created a pytest fixture that checks at the end of the test that no ERROR|CRITICAL logs have been thrown


Alexis-Jacob (approved):
Same thing than Alexandre :D

Issue #168, opened 2020-09-08
Installation with Python 3.5 fails


vdel:
In an environment with Python 3.5 (e.g. Ubuntu 16.04)

```
pip3 install deepomatic-cli
```

eventually leads to:

```
ERROR: Package 'zipp' requires a different Python: 3.5.2 not in '>=3.6'
```

The problem is that we have a dependency `zipp>=0.5` and that apparently the latest version requires Python 3.6.

Running `pip3 install zipp==0.5` prior to running `pip3 install deepomatic-cli` solves the problem.

maingoh:
We don't have any direct requirement for zipp, do you know which library requires zipp ?

vdel:
> do you know which library requires zipp ?

Eeeeh.... no.

maingoh:
Python 3.5 is a bit too old, please upgrade your python installation and feel free to reopen if the issue occurs again :)

maingoh:
Close by #215

PR #215, opened 2021-11-10
Removing support for python < 3.6. Adding support for 3.10.

### LINKED ISSUE #184: Update GitPython
We got some trouble with GitPython 2.1.15 https://github.com/gitpython-developers/GitPython/issues/1030 (the issue talk about 3.1.4 though.. weird)
We kept it to support python2 but we will drop it for the next release so we can update GitPython.

maingoh:
Close #184 #168 

Python 2.7 and 3.5 are not maintained anymore and too old. Now officially supporting python 3.10.

- Split install.sh into multiple install scripts (hoping it wouldn't rebuild when one part of the file was touched, but it wasn't the case ..). The content of them didn't change.
- Bump dependencies


requirements.txt
```diff
@@ -1,12 +1,11 @@
...
+text-unidecode==1.3
 greenlet>=0.4.17,<2
 gevent>=20.9.0,<22.1
-jsonschema==3.0.1
-deepomatic-api==0.9.3
+jsonschema==3.2.0
+deepomatic-api==0.9.4
 PyYAML>=3.12
-GitPython==2.1.15
+GitPython>=3.1.0,<3.2.0
```
alemoreau, line 10: GitPython>=3.1.0 requires python>=3.7, it may cause problems with python3.6 ?

alemoreau (approved):

ibreschi (approved):
Good for me!
Maybe add a line in che pr comment that you added support for 3.10

requirements.txt
maingoh: According to the documentation they changed the requirements only since 3.1.22 so it should be fine ! We should add tests on this part one day :smile:
https://gitpython.readthedocs.io/en/3.1.20/intro.html
https://gitpython.readthedocs.io/en/3.1.22/intro.html

PR #220, opened 2022-01-05
Dev 213 improve output

### LINKED ISSUE #213: Feature request: deepo platform engage-app create print engage app id and drive app id
Airtable url: [https://api.airtable.com/v0/appgWOl8ecHRBewEz/Bugs%20%26%20Issues/rec1KusgBToaToNLT](https://airtable.com/tblL2bTpvN3KcH8BB/viwxchJz20QpypLBP/rec1KusgBToaToNLT)
It would be very useful that when we do a \`deepo platform engage-app create\` the return value contains both the drive and engage app ids. 

Basically you want to do:
```
% deepo platform engage-app create ...
New Engage App created with id: AAA . New Drive App created with id: BBB
% echo $?
AAA BBB
```

This allows to reuse that value easily in subsequent commands (in a bash script, typically).

I use that to create a script that duplicates an entire known workflow (and views, and models, etc.) in any new organization.

### LINKED ISSUE #218: Ease of command output validation for test
While testing we have to parse each command output to:
* validate a command is working (matching message output)
* retrieve returned value such as app_id

While the first point could be achieve by other means (error_code, Exception, ....), the second make test writing tedious and hard to factorize.

```
# We either split and strip to match message substring
result = call_deepo(args)                                                                                                                                                             
message, app_id = result.split(':')                                                                                                                                                   
app_id = app_id.strip()                                                                                                                                                               
assert message == 'New app created with id'
...  
# Or we use slicing ...
result = call_deepo(args)
engage_app_version_id = result[-36:]
assert result[0:18] == 'New app version \'v'
....
```

Instead what we would like is to manipulate an abstract response object (dict, dataclass, ...). It will allow to take care of both point listed before (validate a command is working and extract value of interest):
```
>>> call_deepo(args)
{
    "return_code": 0,
    "app_id": ....
    "version_id": ....
}
```
Todo:
* add at least one mechanism to easily access value from message output. #213 could do the trick or we could add a `--output-format json` option to deepocli. If the second solution is favored, a new separate issue should be open and submit to PM.
* implement the abstract response object (whatever it might be).
* refactor tests to use it.

oquerne:
close #213
close #218

Improve Platform related command output:
* standardize operation `[operation] <resource_name>: [<field_name>=<field_value>, ...]`
* add `--json-output` and  `--json-output-indent`option to output json instead of messages (easier scripting). 
* adapt tests 

Site related command will be updated in an other PR


deepomatic/cli/lib/platform.py
```diff
@@ -26,7 +27,10 @@ def create_app(self, name, description, services):
         }
         ret = self.drive_client.post('/apps', data=data_app)
         app_id = ret['id']
-        return "DriveApp created with id: {}".format(app_id)
+        return PlatformCommandResult(
```
maingoh, line 30: If we want to standardize the formatting we could have an intermediate class which represent an operation like this (not tested):
```python
    class CommandOperation:                                                                                                                                                                                       
        def __init__(self, operation, resource_name, data, important_fields=None, extra_str=''):                                                                                                                                
            if important_fields is None:                                                                                                                                                                          
                important_fields = ['id']                                                                                                                                                                         
            self.operation = operation                                                                                                                                                                            
            self.resource_name = resource_name                                                                                                                                                                    
            self.data = data                                                                                                                                                                                      
            self.important_fields = important_fields
            self.extra_str = extra_str                                                                                                                                                              
                                                                                                                                                                                                                  
       def __repr__(self):                                                                                                                                                                                       
           return str(self)                                                                                                                                                                                      
                                                                                                                                                                                                                 
       def __str__(self):                                                                                                                                                                                        
           important_data = []                                                                                                                                                                                   
           for field in self.important_fields:                                                                                                                                                                   
               important_data.append('{}={}'.format(field, self.data[field]))                                                                                                                                    
           return '[{}] {}: {}'.format(self.operation, self.resource_name, ' '.join(important_data), self.extra_str)                                                                                                             
                                                                                                                                                                                                                 
       def to_json_str(self, *dump_args, **dump_kwargs):                                                                                                                                                         
           return json.dumps(self.data, *dump_args, **dump_kwargs)                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
```

That you could declare like this:
```python
CommandResult([
    CommandOperation('created', 'engage_app', response, ['id', 'drive_app_id'], '(from=123)'),
    CommandOperation('deleted', 'engage_app', response, ['id']),
    ...
])
```

deepomatic/cli/lib/platform.py
```diff
@@ -186,8 +220,16 @@ def create_app_version_from(self,
             for file in files.values():
                 file.close()
 
-        return "EngageApp version created with id: {} from {}. DriveApp version id: {}".format(
-            response['id'],
-            origin,
-            response['drive_app_version_id']
+        return PlatformCommandResult(
+            [
+                "[created] engage_app_version_id ({major}.{minor}): {engage_app_version_id} (from: {origin})",
```
maingoh, line 225: ahah this origin is not really part of the response, not sure what we want to do (display or not) maybe only in the default output (with parenthesis like you did)

maingoh (commented):
Much cleaner :)

deepomatic/cli/lib/platform.py
oquerne: Yes, was not sure about keeping it ^^
Maybe something like this: `[created]...from=123`

deepomatic/cli/cmds/utils.py
```diff
@@ -3,16 +3,39 @@
...
 from deepomatic.cli.common import (SUPPORTED_IMAGE_INPUT_FORMAT, SUPPORTED_IMAGE_OUTPUT_FORMAT,
                                    SUPPORTED_PROTOCOLS_INPUT, SUPPORTED_VIDEO_INPUT_FORMAT,
                                    SUPPORTED_VIDEO_OUTPUT_FORMAT, SUPPORTED_FOURCC,
                                    SUPPORTED_VIDEO_OUTPUT_COLOR_SPACE)
 
+
 logger = logging.getLogger(__name__)
 
 
+@dataclass
```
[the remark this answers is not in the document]
oquerne: Would that be ok for you to add as dependencie https://pypi.org/project/dataclasses/ ?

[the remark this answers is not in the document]
maingoh: Yes you can ! You might need to install it only if python < 3.7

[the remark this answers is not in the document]
maingoh: https://www.python.org/dev/peps/pep-0508/#environment-markers

deepomatic/cli/cli_parser.py
```diff
@@ -53,6 +54,11 @@ def run(args):
     logging.basicConfig(level=log_level, format=log_format)
 
     result = args.func(vars(args))
+
     if result:
-        print(result)
+        if isinstance(result, CommandResult) and args.json_output:
```
maingoh, line 59: nit: maybe for some commands we will have to return a list, but we can change later if needed

maingoh (approved):
Very nice !!

deepomatic/cli/cmds/utils.py
```diff
@@ -10,9 +10,49 @@
...
+    Attributes:
+        resource_name (str)
+        extra (str)
+        operation (str)
+        fields_filter (list)
+        data (dict)
+    """
+    def __init__(self, operation, resource_name, data, fields_filter=None, extra=None):
+        self.operation = operation
+        self.resource_name = resource_name
```
ibreschi, line 29: nit. I would invert the order of this two in the constructor. 
This way it would read 
`        return CommandResult("drive_app", "updated", ret)`
But it's a pure personal taste it's not necessary to do it

ibreschi (approved):
Super work!!!

deepomatic/cli/cmds/utils.py
oquerne: I took a bit of time to think about it and I will stick to the output pattern: "[operation] resource_name: data" order ^^

PR #225, opened 2022-01-31
Add support for .m3u8 video type

romyli:
Adding .m3u8 in the allowed video extensions


maingoh (approved):
We might have some surprise with the fps and frame count depending on what we do because of the stream vs video detection which rely on the input protocol. 
The current logic is:
```
http://fqdm.com/playlist.m3u => deepocli will detect it as stream
```
```
/tmp/playlist.m3u => deepocli will detect as video
```
Which doesn't rely on the m3u content (you could have a stream inside or a video or both).

ibreschi (approved):
.

PR #229, opened 2022-08-04
Add studio json format as inference input

alemoreau:
Studio format (*.txt) was documented and appeared in the `--input` helper, but was not implemented.

### DIFF
--- a/deepomatic/cli/input_data.py
+++ b/deepomatic/cli/input_data.py
@@ -169,6 +169,8 @@ class StudioInputData(InputData):
                                     if os.path.exists(path):
                                         frame = cv2.imread(path)
                                         break
+                                else:
+                                    raise FileNotFoundError(path)
                             elif "url" in image_data:
                                 req = urllib.request.urlopen(image_data.get("url"))
                                 arr = np.asarray(bytearray(req.read()), dtype=np.uint8)
--- a/deepomatic/cli/input_data.py
+++ b/deepomatic/cli/input_data.py
@@ -153,7 +153,7 @@ class StudioInputData(InputData):
     def __iter__(self):
         self._frames = []
         with open(self._descriptor) as f:
-            for line_i, line in enumerate(f.readlines()):
+            for line_i, line in enumerate(f):
                 line = line.strip()
                 try:
                     json_data = json.loads(line)
--- a/deepomatic/cli/input_data.py
+++ b/deepomatic/cli/input_data.py
@@ -3,7 +3,8 @@ import cv2
 import logging
 
 from .common import (SUPPORTED_IMAGE_INPUT_FORMAT, SUPPORTED_PROTOCOLS_INPUT,
-                     SUPPORTED_VIDEO_INPUT_FORMAT, clear_queue)
+                     SUPPORTED_VIDEO_INPUT_FORMAT, SUPPORTED_STUDIO_INPUT_FORMAT,
+                     clear_queue)
 from .exceptions import DeepoFPSError, DeepoInputError, DeepoVideoOpenError
 from .frame import Frame
 from .thread_base import Thread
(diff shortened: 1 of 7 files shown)


deepomatic/cli/input_data.py
```diff
@@ -129,6 +137,68 @@ def is_infinite(self):
...
+        super(StudioInputData, self).__init__(descriptor, **kwargs)
+        self._frames = []
+        self._studio_file_dir = os.path.dirname(self._descriptor)
+        self._name = 'studio_%s_%s' % ('%05d', self._reco)
+        self._iterator = None
+    
+    def __iter__(self):
+        self._frames = []
+        with open(self._descriptor) as f:
+            for line_i, line in enumerate(f.readlines()):
```
maingoh, line 156: Avoid loading the file content in memory

```suggestion
            for line_i, line in enumerate(f):
```

deepomatic/cli/input_data.py
```diff
@@ -129,6 +137,68 @@ def is_infinite(self):
         return False
 
 
+class StudioInputData(InputData):
+    @classmethod
+    def is_valid(cls, descriptor):
+        _, ext = os.path.splitext(descriptor)
+        return os.path.exists(descriptor) and ext.lower() in SUPPORTED_STUDIO_INPUT_FORMAT
```
maingoh, line 144: Small thought on our `--input` option, we currently try to determine the format of each input. It works and is easy when relying on distinct file extensions.
If we start having other formats within the same file extension (.txt) it will be difficult to detect it automatically except by trying to validate the file against all different type of formats. And if doesn't match any format we will have to display all the detections that failed because we will not know which file content the user is trying to send. 
So maybe we should add some explicitness somewhere so that we know that the user really want to send a studio file format. It could be for example `.studio.txt` or `.studio.jsonl`. This is not blocking though, just for the future.

maingoh (approved):

deepomatic/cli/input_data.py
```diff
@@ -129,6 +137,68 @@ def is_infinite(self):
...
+                        images_data = json_data["data"]
+                        for image_data in images_data:
+                            if "file" in image_data:
+                                p = image_data.get("file")
+                                for path in [p, os.path.join(self._studio_file_dir, p), os.path.abspath(p)]:
+                                    if os.path.exists(path):
+                                        frame = cv2.imread(path)
+                                        break
+                            elif "url" in image_data:
+                                req = urllib.request.urlopen(image_data.get("url"))
```
maingoh, line 173: It might be faster to send the url as is to the worker (which is optimized for concurrent download)

deepomatic/cli/input_data.py
alemoreau: I like the .studio prefix, maybe we should add this later when we'll need to support multiple .txt / .jsonl

deepomatic/cli/input_data.py
alemoreau: We probably need the image anyway in deepocli for the outputs (like video outputs)

maingoh: Oh true ! Well we could optimize this to do some concurrent download but it can be done later :)

deepomatic/cli/input_data.py
```diff
@@ -129,6 +137,68 @@ def is_infinite(self):
...
+                        pass
+                    elif schema_type == JSONSchemaType.STUDIO_INPUT:
+                        images_data = json_data["data"]
+                        for image_data in images_data:
+                            if "file" in image_data:
+                                p = image_data.get("file")
+                                for path in [p, os.path.join(self._studio_file_dir, p), os.path.abspath(p)]:
+                                    if os.path.exists(path):
+                                        frame = cv2.imread(path)
+                                        break
```
romyli, line 171: The case where the local path does not exist is not handled, it returns: `Error line 534: local variable 'frame' referenced before assignment`

PR #235, opened 2022-11-08
Fix engage request timeout increase

oquerne:


deepomatic/cli/lib/platform.py
```diff
@@ -13,11 +13,14 @@
 
 
 LOGGER = logging.getLogger(__name__)
+REQUEST_TIMEOUT = float(os.getenv("REQUEST_TIMEOUT", "40."))
```
maingoh, line 16: nit: maybe add an `S` at request to remind it is related to the `requests` package

deepomatic/cli/common.py
```diff
@@ -19,6 +19,7 @@
 SUPPORTED_PROTOCOLS_INPUT = ['rtsp', 'http', 'https']
 SUPPORTED_IMAGE_OUTPUT_FORMAT = SUPPORTED_IMAGE_INPUT_FORMAT
 SUPPORTED_VIDEO_OUTPUT_FORMAT = ['.avi', '.mp4']
+REQUESTS_DEFAULT_TIMEOUT = float(os.getenv("REQUEST_TIMEOUT", "40."))
```
maingoh, line 22:
```suggestion
REQUESTS_DEFAULT_TIMEOUT = float(os.getenv("REQUESTS_TIMEOUT", "40."))
```

deepomatic/cli/lib/platform.py
```diff
@@ -16,8 +17,10 @@
 
 
 class DrivePlatformManager(object):
-    def __init__(self, client_cls=HTTPHelper):
-        self.drive_client = client_cls()
+    def __init__(self):
+        self.drive_client = HTTPHelper(
+            requests_timeout=REQUESTS_DEFAULT_TIMEOUT
```
maingoh, line 22: Can you take the opportunity to add the user agent in all those places as well

tests/test_site.py
```diff
@@ -140,7 +140,8 @@ def tmp():
 @contextmanager
 def setup():
     tmp_dir = tempfile.mkdtemp()
-    manager = SiteManager(path=tmp_dir, client_cls=MockApi)
+    manager = SiteManager(path=tmp_dir)
+    manager._client = MockApi()
```
maingoh, line 144: Actually maybe for the site_manager we should keep the `client_cls` (if the HTTPHelper fails to instanciate before patching it)

maingoh (approved):
Nice !!

tests/test_site.py
maingoh: Or use `unittest.patch` to replace `HTTPHelper` class with `MockApi`

alemoreau (approved):

